PR #4, opened 2024-07-30
Measure benchmark runtime and memory consumption

jacek-oet:


siddharth-krishna (changes requested):
Thanks, on the whole this looks really good! Some requests:

- [ ] Let's not check in the NC files to the repo, they will start getting too large. For now, could you put them in some public web archive (e.g. your OET google drive, or feel free to suggest better alternative), and have the script download them given the URLs?
  - [ ] Also, please convert it from Pypsa network to Linopy model and then upload them.
- [x] Let's do a separate run to measure runtime and a separate run to measure memory, because memory profiling will introduce a runtime overhead. Let's report this overhead to stdout
- [ ] Let's discuss how we can test the accuracy of the memory measurement, and see if we need to use some other tool to do it. Some ideas:
  - [x] Look into https://github.com/PyPSA/pypsa-eur/blob/master/scripts/_benchmark.py and see how it works and how it is used
  - [x] Let's test that tracemalloc also includes memory consumption of the solver (which is usually written in C). 
    - [x] One idea to do this: compare tracemalloc result to `/usr/bin/time -v`. (First look up if `/usr/bin/time` is accurate in WSL?)
    - [x] Alternative: compare the result of a normal run and a run where the `linopy.solve()` method is commented out.
  - [x] Let's use the linopy memory test [benchmark model](https://github.com/PyPSA/linopy/blob/master/linopy/examples.py#L55) to check that the reported memory increases with model size

jacek-oet:
@siddharth-krishna 
**Re: Look into https://github.com/PyPSA/pypsa-eur/blob/master/scripts/_benchmark.py and see how it works and how it is used**
I tried to use function memory_logger of PyPSA _benchmark, it always return bigger memory value than me get_traced_memory of tracemalloc so I think tracemalloc doesn't includes memory consumption of the solver (which is usually written in C).
1. **Memory\_logger (memory\_profiler)**:
    - `memory_logger` uses `memory_profiler`, which relies on querying the process's memory usage at regular intervals.
    - It is a pure python module which depends on the [psutil](http://pypi.python.org/pypi/psutil) module.
2. **Tracemalloc**:
    - `tracemalloc` focuses on tracing memory allocations in Python, which might exclude some memory usages such as those by external libraries or subprocesses.
    - Measures memory allocated by Python objects, which may not account for memory used by non-Python objects or overhead not tracked by the Python interpreter.

---

**Re: Let's test that tracemalloc also includes memory consumption of the solver (which is usually written in C). One idea to do this: compare tracemalloc result to `/usr/bin/time -v`. Alternative: compare the result of a normal run and a run where the `linopy.solve()` method is commented out.**
If linopy.solve() method is commented out, memory usage calculated by tracemalloc return 0

---

**Re: Let's use the linopy memory test [benchmark model](https://github.com/PyPSA/linopy/blob/master/linopy/examples.py#L55) to check that the reported memory increases with model size**

Can you give me more detail, is that we get model size;
```
linopy_model_path = 'runner/' + file_path.replace(".nc", "-linopy.nc")
os.path.getsize(linopy_model_path)
```
and Memory Usage compare with other model with different size ?

siddharth-krishna:
> I tried to use function memory_logger of PyPSA _benchmark, it always return bigger memory value than me get_traced_memory of tracemalloc

Could you please document here how you ran this test (which commit, what command to run, etc), and the results (memory values reported for around 3 benchmarks), so that others can also review and reproduce if necessary? Thanks

>  If linopy.solve() method is commented out, memory usage calculated by tracemalloc return 0

Did you comment out the line here? I guess then there's no code really between the tracemalloc start and stop!
image

I guess a better way to do this test is to modify the linopy code and comment out the line that actually calls the solver?
https://github.com/PyPSA/linopy/blob/915e7a2c642ae435228a9cc2f31e858bf1c235aa/linopy/model.py#L1078
(You can install a local checkout of linopy in your virtualenv by going to the linopy folder and `pip install -e .`)

> Re: Let's use the linopy memory test [benchmark model](https://github.com/PyPSA/linopy/blob/master/linopy/examples.py#L55) to check that the reported memory increases with model size

What I was thinking is to write a script that calls `m = linopy.examples.benchmark_model(n)` for `n` in `[10, 100, 1000, etc]` and measures the memory usage of `m.solve()` using the different techniques we have so far, to confirm that all techniques report increasing amount of memory used as `n` increases.

Btw, if you write standalone scripts for experiments like the above, you can add it to this PR at `pocs/measure-memory/` or something like that, so that we retain the scripts for the record. Thanks very much!

website/compare.py
```diff
@@ -29,23 +29,23 @@
...
+solver2 = st.selectbox("Select Solver 2", df["Solver"].unique())
 
 if st.button("Compare Solvers"):
     # Filter data for the selected solvers
-    solver1_data = df[df["Solver Name"] == solver1]
-    solver2_data = df[df["Solver Name"] == solver2]
-
+    solver1_data = df[df["Solver"] == solver1]
+    solver2_data = df[df["Solver"] == solver2]
+    print('solver1_data', solver1_data, 'solver2_data', solver2_data, 'solver1', solver1, 'solver2', solver2)
```
siddharth-krishna, line 48: Is this a debug print that's leftover?

website/compare.py
```diff
@@ -29,23 +29,23 @@
     unsafe_allow_html=True
 )
 
-data_url = "./pocs/solvers.csv"
+data_url = "./pocs/benchmark_results.csv"
```
siddharth-krishna, line 32: Is the main results file this, or `pocs/benchmark.csv`?

siddharth-krishna (commented):
Thanks, Jacek, this all looks good so far. I left some minor suggestions.

So to confirm my understanding: `tracemalloc` might not be counting the memory consumption of the solver code in C, because it reports lower numbers than pypsa's `_benchmark.py`.

If so, I think the next step is to validate the numbers reported by pypsa's `_benchmark.py` against some system tool like `/usr/bin/time -v` on linux. Did you check if this tool is accurate on WSL? If not, we can provision you a GCP linux VM to do these tests.

Then, the last TODOs before merging this PR:
- [x] Get the NC files from GDrive or something instead of checking it into the repo
- [ ] Use the final decision method (e.g. pypsa's `_benchmark.py`) in the `runner/app.py` (could we rename it `run_benchmarks.py` btw?)

pocs/measure-memory/REAMDE.md
```diff
@@ -0,0 +1,38 @@
+## Comparing Memory Usage: PyPSA Benchmark's memory_logger vs. tracemalloc.get_traced_memory
```
siddharth-krishna, line 1: Thanks for adding these here, please could you also remove it from the main README?

jacek-oet: Updated.

jacek-oet:
@siddharth-krishna 
I updated code, so now we only use benchmark_results.csv in pocs folder.
I also removed NC files in pocs/measure-memory/.

Issue #6, opened 2024-08-06
Website: first quarter version


danielelerede-oet:
- [x] Can we use the `plotly` library to generate all the plots/charts on the website, so that they are interactive?
- [x] Compare Solvers page:
  - [x] The first chart (scatter plot): use the same scale for both axes and add a line at `y=x` so that it is easy to see the benchmarks for which one solver is faster than the other. (See below for example.)
  - [x] Add a similar scatter plot that compares the peak memory consumption of the two selected solvers
  - [x] Can the default values for the dropdowns be Gurobi vs Highs, instead of both being equal?
  - [x] When hovering on a datapoint in both plots, please display the benchmark name too
- [x] Solver history page:
  - [x] The chart should have "solver version" in the x-axis, "runtime" in the y-axis, and should have one line per benchmark so that it is possible to see whether the solver is getting faster or slower on the benchmarks
  - [x] Add a similar plot that compares the peak memory consumption of the different solver versions
- [x] Add a new Home page:
  - [x] Add a plot to the top of the page that plots "runtime" on the x-axis, "peak memory" on the y-axis, has one point per benchmark (with all benchmarks connected by a line) and one color per solver. This plot should have all the data that we have for every benchmark and every solver right now. When we extend to having multiple versions of each solver, this plot should only plot the latest version of each solver.
  - [x] Can we add a second plot to this page, similar to the one above, but where the lines connect the points for all solvers but for a particular benchmark?
- [x] Add a new Benchmarks page:
  - [x] This page has a dropdown to select the benchmark, and upon selection it shows a table of metadata about each benchmark. The metadata is in information in [this](https://github.com/open-energy-transition/solver-benchmark/blob/main/benchmarks/pypsa/README.md#description-of-the-sample-problems) table. I suggest moving the information from that table to a YAML file in that same folder, and then reading the YAML file in the website to display on this new page.
  - [x] Upon selection of a benchmark, along with the metadata table, also show the "peak memory vs runtime" plot from the Home page, but filtered to only the data points corresponding to this benchmark.
- [x] Round all runtimes to 1 decimal place, memory to 0 decimal places, objective values to 2 decimal places
- [x] add a sensible title on the landing page like ("OET's super trooper duper awesome benchmark" 😉 )
- [x] add a welcoming text on the landing page (context: what is it, why is it important, who is doing/funding)
[...]

PR #16, opened 2024-08-28
[website] Add home page; use plotly for charts

jacek-oet:
See https://github.com/open-energy-transition/solver-benchmark/issues/6


website/compare.py
```diff
@@ -23,6 +23,23 @@
...
     solver2_data = df[df["Solver"] == solver2]
     # Create the scatter plot
-    fig = create_comparison_chart(solver1_data, solver2_data, solver1, solver2)
+    run_time_fig = create_comparison_chart(
+        solver1_data,
+        solver2_data,
+        solver1,
+        solver2,
+        metric_name="Runtime (s)",
+        comparison_type="Runtime"
```
siddharth-krishna, line 32:
```suggestion
        metric_name="runtime (s)",
        comparison_type="runtime"
```

website/compare.py
```diff
@@ -23,6 +23,23 @@
...
+    )
 
-    st.pyplot(fig)
+    mem_use_fig = create_comparison_chart(
+        solver1_data,
+        solver2_data,
+        solver1,
+        solver2,
+        metric_name="Memory Usage (MB)",
+        comparison_type="Memory Usage"
```
siddharth-krishna, line 41:
```suggestion
        metric_name="peak memory usage (MB)",
        comparison_type="peak memory usage"
```

siddharth-krishna: Also please let's double check: is it MB or MiB? (I always forget the difference 😅 )

jacek-oet:
The result of /usr/bin/time -v includes the line:

`Maximum resident set size (kbytes): x
`

However, the unit kbytes is not explicitly mentioned in the [Ubuntu Units Policy](https://wiki.ubuntu.com/UnitsPolicy). I believe this should be interpreted as kB because:

1 kB = 1,000 bytes (Note: small "k")
I found three sources that also mention this as referring to kilobytes.


> https://linux.die.net/man/2/getrusage
> https://man7.org/linux/man-pages/man2/getrusage.2.html
> *This is the maximum resident set size used (in kilobytes).

> *ru_maxrss    the maximum resident set size utilized (in kilobytes).
> https://manpages.ubuntu.com/manpages/bionic/man2/getrusage.2freebsd.html?_gl=1*kksf9u*_gcl_au*MTM2NzU0OTY2OC4xNzI0OTgxMzMw&_ga=2.180816673.782637710.1724978835-620433090.1724978835

siddharth-krishna (approved):
Thanks, this is great!

PR #22, opened 2024-09-09
More runner improvements and instructions to reproduce PyPSA benchmarks

siddharth-krishna:
This PR:
- Changes the pypsa benchmarks:
  - To use a new linopy flag `only_generate_problem_files` instead of `keep_files` as recommended in a discussion with Fabian. This is because we need NC files for benchmarking, not LP, as the latter cannot be read as input to linopy.
  - Refines the instructions on how to generate NC files, in a way that can be automated by CI in the future.
- Improves the benchmark runner:
  - Reverts to measuring time inside `run_solver.py` in order to exclude the sometimes large time taken to `import linopy`
  - Uses `/usr/bin/time --format` to simplify parsing the memory consumption
  - Also records solver status, termination condition, and objective value
  - Checks that `run_solver.py` actually ran without errors, to prevent silent failures
  - Adds a timeout feature, and handles times out and solver errors gracefully


benchmarks/pypsa/README.md
```diff
@@ -1,5 +1,78 @@
...
+     ```bash
+     micromamba env create -f envs/environment.yaml
+     micromamba activate pypsa-eur
+     pip install git+https://github.com/open-energy-transition/linopy.git@only-generate-problem-files --no-deps
+     ```
+
+1. Run the following commands to generate the NC files for each benchmark. This assumes your current working directory is `pypsa-eur/` and that this repo is checked out at `~/code/solver-benchmarks/`; modify this path accordingly.
+     ```bash
+     time snakemake -call all --cores all --printshellcmds --configfile  ~/code/solver-benchmark/benchmarks/pypsa/pypsa-eur-sec-2-lv1-3h.yaml ; echo -e '\a'
+     time snakemake -call solve_elec_networks --cores all --printshellcmds --configfile  ~/code/solver-benchmark/benchmarks/pypsa/pypsa-eur-elec-20-lvopt-3h.yaml ; echo -e '\a'
```
danielelerede-oet, line 17: If we're actually going with the network clustered to 10 nodes `pypsa-eur-elec-20-lvopt-3h.yaml` should be changed to `pypsa-eur-elec-10-lvopt-3h.yaml``

benchmarks/pypsa/pypsa-eur-elec-20-lv1-3h-op.yaml
danielelerede-oet, line 1: The name of the file should be changed to `benchmarks/pypsa/pypsa-eur-elec-10-lv1-3h-op.yaml`

danielelerede-oet (changes requested):
The only change needed from my side regard the name given to the sample sample problem `pypsa-eur-elec-20-lv1-3h-op.yaml` as now it's clustered to 10 nodes and not 20 anymore

jacek-oet (approved):
LGTM

danielelerede-oet (approved):

PR #23, opened 2024-09-10
Handle timeouts in the website

jacek-oet:


website/home.py
```diff
@@ -60,3 +75,12 @@
 
 # Show the plot
 st.plotly_chart(fig)
+
+# Add a line of text explaining the symbols
+st.markdown(
+    """
+    **Legend Explanation:**
+    - **X**: Timeout (TO)
+    - **Circle**: Successful run (OK)
```
siddharth-krishna, line 84:
```suggestion
    - **O**: Successful run (OK)
```

siddharth-krishna (approved):

website/home.py
jacek-oet: I applied.

PR #24, opened 2024-09-11
Add benchmarking metrics and methodology

siddharth-krishna:


docs/Metrics_and_methodology.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Metrics
+
+We record the following metrics for each benchmark and solver combination:
+
+1. **Runtime**: of the `linopy.Model.solve()` call to the solver, mean over `N` iterations
+1. **Peak memory consumption**: of the script `runner/run_solver.py` that uses linopy to call the solver, as reported by `/usr/bin/time -f %M`
+1. **Status**: OK, TO (timeout), ER (error)
+1. **Termination condition**: as returned by the solver, for example optimal, infeasible, unbounded, …
+1. **Objective value**: (a floating point value) the optimal objective value
```
danielelerede-oet, line 11: Hi Sid, not sure if the objective function value could be an interesting metric to report, unless we consider getting exactly the same benchmarks as the ones modelled with PyPSA (but that's something I wouldn't hope for so much)

siddharth-krishna: Thanks, Daniele! Yes, I was not sure about this one either, but someone told me it's good to at least make a record of, for instance to debug a particular solver. We probably won't show it on graphs/charts on the website but just make it available in the full results for any interested users.

danielelerede-oet: Yes, I remember we mentioned that during one of the last meetings. And it's also ok to just keep track of it :)

danielelerede-oet (approved):

PR #25, opened 2024-09-11
[website] A new Home plot, better tooltips, round runtime & memory

jacek-oet:


website/raw-results.py
```diff
@@ -21,6 +24,10 @@
 data_url = Path(__file__).parent.parent / "results/benchmark_results.csv"
 
 df = pd.read_csv(data_url)
+# Loop through each row and update the 'Salary' column by adding 1000
+for index, row in df.iterrows():
+    df.at[index, "Objective Value"] = round_number(row["Objective Value"], 2)
+    df.at[index, "Runtime (s)"] = round_number(row["Runtime (s)"], 1)
+    df.at[index, "Memory Usage (MB)"] = round_number(row["Memory Usage (MB)"], 0)
```
siddharth-krishna, line 31: It seems inefficient to go through every cell in the DF and round it. Please can you look up whether the data frame has an option to round all values in a column to a given precision?

siddharth-krishna (approved):
Great, thank you!

PR #27, opened 2024-09-12
Second benchmark run with T=15min; add some descriptive text to website

siddharth-krishna:


website/history.py
```diff
@@ -93,10 +93,10 @@
 
 st.markdown(
     """
-        **Legend Explanation:**
-        - **X**: Timeout (TO)
-        - **O**: Successful run (OK)
-        """
+    These plots show the evolution of solver performance over time.
+
+    **Legend:** an **X** represents benchmarks that timed out (TO), while an **O** indicates a successful run (OK).
```
danielelerede-oet, line 98: maybe $\bullet$ instead of **O**?

website/home.py
danielelerede-oet, line 1: The homepage looks great now. It would be just nice to have all the legends aligned (so the same width for each plot)

PR #28, opened 2024-09-13
[website] Adjust width, better legend

siddharth-krishna:


website/home.py
```diff
@@ -5,22 +5,21 @@
...
+# st.markdown(
+#     """
+#     <style>
+#     .block-container {
+#       max-width: 100%;
+#     }
+#     </style>
+#     """,
+#     unsafe_allow_html=True,
+# )
```
jacek-oet, line 22: Can we just remove?

jacek-oet (approved):
LGTM

PR #29, opened 2024-09-13
[website] Interactive scatter plot and benchmark metadata on home page

jacek-oet:


siddharth-krishna:
@jacek-oet I played around with this `on_select` feature and figured out how to make a simple scatter plot where if you select a point, it highlights all other points that have the same `solver` by increasing their marker size:
```python
data = pd.DataFrame({'x': [1, 1, 2, 2], 'y': [1, 2, 1, 2], 'solver': ['gurobi', 'highs', 'highs', 'gurobi']})

try:
    print(st.session_state.my_scatter)
    selected_solver = data.at[st.session_state.my_scatter.selection['point_indices'][0], 'solver']
except AttributeError:
    selected_solver = None
marker_sizes = [15 if solver == selected_solver else 8 for solver in data["solver"]]

fig = go.Figure()

fig.add_trace(go.Scatter(
    x=data['x'],
    y=data['y'],
    mode='markers',
    marker=dict(
        size=marker_sizes,
        color=[1 if solver == 'gurobi' else 2 for solver in data['solver']],
        colorscale=['blue', 'green'],
        showscale=False
    ),
))
st.plotly_chart(fig, on_select="rerun", selection_mode='points', key='my_scatter')
```

If it works for you as well, I think we can use this technique to add the desired interactive feature to the scatter plot on the home page. Just remember, what we want in that case is when we click a point, we highlight all other points with the same benchmark name, not solver as in my example above. Also, instead of just making the marker bigger, perhaps we can change the color of all other markers so that they appear greyed out.

jacek-oet:
> @jacek-oet I played around with this `on_select` feature and figured out how to make a simple scatter plot where if you select a point, it highlights all other points that have the same `solver` by increasing their marker size:
> 
> ```python
> data = pd.DataFrame({'x': [1, 1, 2, 2], 'y': [1, 2, 1, 2], 'solver': ['gurobi', 'highs', 'highs', 'gurobi']})
> 
> try:
>     print(st.session_state.my_scatter)
>     selected_solver = data.at[st.session_state.my_scatter.selection['point_indices'][0], 'solver']
> except AttributeError:
>     selected_solver = None
> marker_sizes = [15 if solver == selected_solver else 8 for solver in data["solver"]]
> 
> fig = go.Figure()
> 
> fig.add_trace(go.Scatter(
>     x=data['x'],
>     y=data['y'],
>     mode='markers',
>     marker=dict(
>         size=marker_sizes,
>         color=[1 if solver == 'gurobi' else 2 for solver in data['solver']],
>         colorscale=['blue', 'green'],
>         showscale=False
>     ),
> ))
> st.plotly_chart(fig, on_select="rerun", selection_mode='points', key='my_scatter')
> ```
> 
> If it works for you as well, I think we can use this technique to add the desired interactive feature to the scatter plot on the home page. Just remember, what we want in that case is when we click a point, we highlight all other points with the same benchmark name, not solver as in my example above. Also, instead of just making the marker bigger, perhaps we can change the color of all other markers so that they appear greyed out.

Sure Got it. I will update.

jacek-oet:
@siddharth-krishna I fixed but.
I’m trying to fix an issue with the legend on the right. It turns grey when a point is selected, and I’m looking for a way to resolve it.

https://github.com/user-attachments/assets/583111c2-9336-434c-a021-69296661b0c7

siddharth-krishna:
Thanks for looking into this! Yes, the legend becoming grey is not ideal. We also want a way to make the selected data points appear in front of the greyed out data points.

If this is looking difficult to achieve, let's not spend too much time on it. Here's an alternative idea:

When clicking on a data point in the graph, let's only display the data points corresponding to the same benchmark in the scatter plot. And let's maybe show the metadata of that benchmark below the scatter plot, instead of the table that shows the metadata for all the benchmarks that we now have above the scatter plot. And add a button called "Show all" or something that allows users to see all data points again.

siddharth-krishna (approved):
Thanks, Jacek! Ideally the assertion should have gone in #37 but it's good here for now.

PR #32, opened 2024-09-16
Update the 2 simple pypsa problems

danielelerede-oet:


danielelerede-oet:
Hi @siddharth-krishna , I updated the two, simple pypsa-based problems adding a gas generator and applying unit commitment on it instead of wind generators (in the last problem only). I also updated the README.md and the metadata file.

siddharth-krishna (changes requested):
Thanks, Daniele! This looks good, except for one minor thing: I don't think the `benchmarks/pypsa/.idea/workspace.xml` file should be checked into the repo. See e.g. https://stackoverflow.com/questions/43198273/which-files-in-idea-folder-should-be-tracked-by-git

Perhaps you can remove the file from this branch, and add a line `.idea/` to the repo's `.gitignore` file to prevent this happening in the future? Thanks!

danielelerede-oet:
Hey @siddharth-krishna , I did as you suggested, plus I noticed the names of the sample problems should have been changed to include the information on the presence of the gas generator.

siddharth-krishna (approved):
Looks good, thank you! :)

PR #33, opened 2024-09-16
Implement Tag-Based Filtering for Benchmark Plots Across Categories

jacek-oet:


website/home.py
```diff
@@ -1,14 +1,27 @@
...
+def load_metadata(file_path):
+    with open(file_path, "r") as file:
+        return yaml.safe_load(file)
+
+
+metadata = load_metadata("benchmarks/pypsa/metadata.yaml")
+
+# Convert metadata to a DataFrame for easier filtering
+metadata_df = pd.DataFrame(metadata).T.reset_index()
+metadata_df.rename(columns={"index": "Model Key"}, inplace=True)
```
siddharth-krishna, line 23:
```suggestion
metadata_df.rename(columns={"index": "Benchmark Name"}, inplace=True)
```

website/home.py
```diff
@@ -31,16 +44,95 @@
...
+    unsafe_allow_html=True,
+)
+
+# Create a sidebar for filters
+with st.sidebar:
+    st.markdown("### Filters")
+
+    # Add select boxes for filtering metadata with default values set to all items
+    selected_model_name = st.multiselect(
+        "Select Model Name",
```
siddharth-krishna, line 72:
```suggestion
        "Model Name",
```
And similarly for the ones below

website/home.py
```diff
@@ -31,16 +44,95 @@
...
+    metadata_df["Sectors"].isin(selected_sectors) if selected_sectors else True
+)
+
+filtered_metadata = metadata_df[
+    mask_model_name & mask_technique & mask_problem_kind & mask_sectors
+]
+
+# Show filtered metadata (optional)
+if not filtered_metadata.empty:
+    st.write("### Benchmark Metadata")
```
siddharth-krishna, line 117:
```suggestion
    st.write("### Benchmark Details")
```

siddharth-krishna (commented):
Looking great so far, thanks! I left some minor suggestions

siddharth-krishna (commented):
Thanks, Jacek! This looks good. Some minor requests:
- [x] I notice the filtering code is repeated for each page. Can we refactor it into a shared module and reuse the code?
- [x] Home page, under `Benchmark Details` header: could we add a sentence saying either `Showing all benchmarks` or `Filters are active; showing X/Y benchmarks`?
- [x] Actually, let's add a sentence like the above in all pages where we have filters.
- [x] Home page: can we move the violin plots below the scatter plot?

Thanks!

siddharth-krishna (approved):
Thanks again, the changes look good. Just a minor requests and then you can merge it:
- [ ] Would you mind moving the "showing all benchmarks"/"filters are active..." message from being a header on the main page, which looks a little awkward, to the sidebar to be a sentence below the filters themselves?

PR #35, opened 2024-09-25
[benchmarks/pypsa] Simplify config files and shrink benchmarks

danielelerede-oet:
Hi @siddharth-krishna , in the last commit I removed those lines that `pypsa-eur-elec-10-lvopt-3h.yaml` has in common with the `common.default.yaml` file for pypsa v 0.11.0 (which is the one this config file refers too). However, apart from reducing the size of the file I'm not so convinced of the reason why we should need to cut parts of this file. I guess we'll be having a reference `config.default.yaml` file and an additional script to modify it according to the specific info related to each of the pypsa-eur-based sample problems, but I can't see why we should be doing that, if so.


siddharth-krishna (approved):
@danielelerede-oet this looks good to me. I think the motivation for removing the common lines with the default config is so that it is easier for humans to look at the different benchmark configs and visually see what each of them is about and how they differ.

I don't think we need to have an additional script that modifies this config and brings in info from the default config -- I think snakemake does that automatically. Could you please check that you are able to build the same model using the reduced config compared to the config on the `main` branch?

danielelerede-oet:
Hi @siddharth-krishna, the modifications in the latest commit comply with the following requirements:
1) modify the metadata as in contrast with Fabian's first classification the problem with transmission expansion is actually not a MILP
2) update the pypsa-eur-based UC problem removing the linearization of UC
3) change the rules for solving the operational problems given the latest updates to pypsa-eur 

Moreover, I'm currently experiencing issues with the `solve_operations_network` rule in pypsa-eur after the latest changes done in the repo. I already opened an issue about that (https://github.com/PyPSA/pypsa-eur/issues/1344).

benchmarks/pypsa/pypsa-eur-elec-20-lv1-3h-op-ucconv.yaml
```diff
@@ -1,1205 +1,306 @@
...
-    links: "#8a1caf"
-    HVDC links: "#8a1caf"
-    DC: "#8a1caf"
-    DC-DC: "#8a1caf"
-    DC link: "#8a1caf"
-    load: "#dd2e23"
-    waste CHP: '#e3d37d'
-    waste CHP CC: '#e3d3ff'
-    HVC to air: 'k'
+      linearized_unit_commitment: false
```
siddharth-krishna, line 16: Shouldn't this line also be under the `options:` key? E.g.:
```suggestion
  options:
      linearized_unit_commitment: false
```

siddharth-krishna: And similarly in the other files. Is there a way we can generate new LP files given these configs, and check that it has MILP features, if you haven't already done this?

benchmarks/pypsa/pypsa-eur-elec-20-lv1-3h-op-ucconv.yaml
```diff
@@ -1,1205 +1,306 @@
...
+
+# docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#sector
+sector:
```
siddharth-krishna, line 19: Am curious why the `sector:` section has been uncommented, compared to the old version of this file?

siddharth-krishna: Btw, a tip for next time, if you don't mind: here it looks like you changed the position of the `sector:` section in this file. This means git can't figure out what lines have changed, so in the "Files changed" tab on GitHub it looks like the section was deleted and added again in a new place. If you had instead retained the position of the section in the file, it might be able to only show the changes to it (most lines have just been uncommented, a few have been added, as far as I can make out), which would make review easier. Thanks!

siddharth-krishna (commented):
Thanks, Daniele! I notice a potential bug with the `linearized_unit_commitment` line, otherwise it looks good. 

I assume the plan is to wait for the pypsa-eur issue to be fixed before we merge in this PR?

benchmarks/pypsa/pypsa-eur-elec-20-lv1-3h-op-ucconv.yaml
danielelerede-oet: Thanks Sid, you're right! That issue should be present only in this file anyway, in the others we don't have unit commitment. I already generated the LP file and checked it for the presence of MILP variables, and I also sent it to Julian, you can find it at https://drive.google.com/file/d/1Y5QTw1wnjaRLDG4x2PZ2MQzamMHcOP_w/view?usp=drive_link

benchmarks/pypsa/pypsa-eur-elec-20-lv1-3h-op-ucconv.yaml
danielelerede-oet: I uncommented it because I noticed that PyPSA would have downloaded all the sector-coupled-related data even though useless for an operational run. 
Thanks for you suggestion anyway, I'll be more careful next time!

danielelerede-oet:
@siddharth-krishna today I updated metadata with continuous and binary variables for MILP problems (I also noticed there was a mistake before because I didn't know how to properly read the number of variables, but Julian gave wonderful suggestions on this) and resolved a conflict in the solution of `pypsa-gas+wind+sol+ely-1h-ucgas.py` (that doesn't change anything at all in the formulation of the problem, it's just that pypsa was raising warnings due to this). I fear anyway that the "only_generate_problem_file" option doesn't work for `pypsa-gas+wind+sol+ely-1h-ucgas.py` and `pypsa-gas+wind+sol+ely-1h.py`

danielelerede-oet:
Hey @siddharth-krishna, I changed the names for pypsa benchmarks according to our discussion and updated metadata. I guess the runner and maybe some other files require some modifications, too related to this. I also removed information about the absence of transmission expansion (formerly indicated through `lv1`) and, when present, I'm now identifying it through `trex` (let me know if you have better ideas) instead of `lvopt` which is very "pypsa-sounding".

siddharth-krishna:
I'm merging this PR now in because we need to build on it for upcoming PRs. Daniele, when you're back next week, could you perhaps review my changes and let me know if you'd like any changes? I can do them in another PR.

My main change was to reduce nodes and increase time resolution to make all the benchmarks as small as possible. This is in order to make all solvers solve them in a short amount of time, which eases our testing. One of my next work items is to create multiple sizes of each benchmark, so the original sizes will be re-added to the benchmark set soon. I hope that's okay.

I notice there's a bug with the GLPK results: all the statuses are `warning` and not `ok`. I'll open an issue to investigate this.

PR #41, opened 2024-10-15
Update the runner to work on LP file

jacek-oet:
### Summary
This pull request updates the runner to work with LP files.

### Result
- Command

`python runner/run_benchmarks.py 
`
- Config: 
```
solvers = ["highs", "glpk"] 
benchmark_files_info: {
            "name": "pypsa-gas+wind+sol+ely-1h.lp",
            "url": "https://drive.usercontent.google.com/download?id=1D0_mo--5r9m46F05hjHpdzGDoV0fbsfd&export=download&authuser=0",
 },

```

- Log:
```
Running solver highs on pypsa-gas+wind+sol+ely-1h.lp (0)...
metrics
{'runtime': 84.55590558052063, 'status': 'ok', 'condition': 'optimal', 'objective': 84667526618.30806, 'memory': 401.868}
Running solver glpk on pypsa-gas+wind+sol+ely-1h.lp (0)...
metrics
{'runtime': 185.00233101844788, 'status': 'ok', 'condition': 'optimal', 'objective': 84667526620.0, 'memory': 234.424}
Benchmarking complete.
```

_benchmark_results.csv_

```
Benchmark,Solver,Status,Termination Condition,Objective Value,Runtime (s),Memory Usage (MB)
pypsa-gas+wind+sol+ely-1h,highs,ok,optimal,84667526618.30806,84.55590558052063,401.868
pypsa-gas+wind+sol+ely-1h,glpk,ok,optimal,84667526620.0,185.00233101844788,234.424

```


runner/run_solver.py
```diff
@@ -17,19 +17,30 @@ def get_solver(solver_name):
     return solver_class()
 
 
+def select_solver_method(solver_name, solver, problem_file):
+    if solver_name == "glpk":
+        solution_fn = f"runner/solution/{problem_file.stem}.sol"
+        return lambda: solver.solve_problem(
+            problem_fn=problem_file, solution_fn=solution_fn
```
siddharth-krishna, line 24: Do we need to special case this for glpk? What happens if we also specify the `solution_fn` flag for other solvers?

README.md
```diff
@@ -2,6 +2,35 @@
...
+1. Install Git using Micromamba:
+   ```shell
+      micromamba install -n base -c conda-forge git -y
+   ```
+2. Clone and install the custom version of Linopy:
+   ```shell
+      git clone https://github.com/daniel-rdt/linopy
+      cd linopy
+      git checkout linopy_solver_class
+      pip install -e . --no-deps
```
siddharth-krishna, line 30: I don't think you need to clone and install a local folder if you don't want to make changes to Daniel's branch. If his branch works without modifications, we should be able to install from his branch directly and specify this in `requirements.txt`. See e.g.
https://github.com/drifter089/solver-benchmark/pull/2/files#diff-7de95f23f030adf370bf4a5a57623b318f3a6caa757870fbb427c9560b703e4a

README.md
```diff
@@ -2,6 +2,35 @@
...
 This repository contains code for benchmarking LP/MILP solvers, and an interactive website for analyzing the results.
 
+
+## Run Custom Linopy by Daniel-RDT
+#### TODO: Remove this section once the pull request for linopy_solver_class by Daniel-RDT is merged into the main Linopy repository.
+### Install micromamba
+1. Update your system and install bzip2:
+   ```shell
+      sudo apt update
+      sudo apt install bzip2 -y
```
siddharth-krishna, line 12:
```suggestion
   ```shell
   sudo apt update
   sudo apt install bzip2 -y
```
Don't put extra spaces in the front of shell commands, otherwise when copying and pasting these, this will also paste the extra spaces:
image

This is bad because Linux shells like bash will remove a command from the shell history if it starts with a space.

runner/run_solver.py
jacek-oet: We can add the solution_fn flag for other solvers.

I have a question: When generating the solution file (.sol), do we need to generate a separate file for each iteration, or will the latest run override the previous solution file? For example, if we have iterations = 2, will we only have one solution file for the last iteration?

README.md
jacek-oet: Got it, I updated this document. We can run the following command directly to install via pip, so we don't need requirements.txt for custom linopy
```shell
pip install git+https://github.com/daniel-rdt/linopy.git@linopy_solver_class
```

runner/run_solver.py
siddharth-krishna: For benchmarking, I don't think we care about the solution files. So it is okay to overwrite them. If we need anything from the solution file, we can use the last iteration's file.

siddharth-krishna (approved):
Thanks! Looks good :)

Issue #14, opened 2024-08-27
Automatically generate benchmark LPs


siddharth-krishna:
Write a script to follow the instructions [here](https://github.com/open-energy-transition/solver-benchmark/blob/main/benchmarks/pypsa/README.md) and automatically generate the LP files ready for benchmarking.

Is there a limit on the amount of CI minutes for a GitHub public repo? 

Some features that are good to have:
- [x] We need a way to tell linopy to generate the NC file in a specific path so that the CI script can use the result. Suggestion: modify the `only_generate_problem_file` config flag (branch: `open-energy-transition/linopy.git@only-generate-problem-files`) to take a file path instead of true/false, and use that path in linopy to dump the file.
  - [x] Also look at the extension and generate NC or LP file accordingly
  - [ ] When this works, open a PR to upstream linopy
- [x] Pypsa-eur's `scripts/solve_network.py` needs to be patched as mentioned in the [readme](https://github.com/open-energy-transition/solver-benchmark/blob/main/benchmarks/pypsa/README.md) to support the `only_generate_problem_file` config flag. Sid will make an upstream PR with this soon, until then the CI needs to patch the python file.
- [x] As reproducible as possible: use fixed commit hashes when checking out repos, use pinned versions when installing python packages, use fixed hashes of docker images if possible

jacek-oet:
@siddharth-krishna I’m stuck at step 3, I run this
```
time snakemake -call all --cores all --printshellcmds --configfile  ~/code/solver-benchmark/benchmarks/pypsa/pypsa-eur-sec-2-lv1-3h.yaml ; echo -e '\a'
```
and it show error
```
Error in rule retrieve_worldbank_urban_population:
    jobid: 41
    output: data/worldbank/API_SP.URB.TOTL.IN.ZS_DS2_en_csv_v2_3403768.csv
```

jacek-oet:
I think our target is to reduce the time it takes to run CI, so we should address these issues first.

**1. Write a script to follow the instructions [here](https://github.com/open-energy-transition/solver-benchmark/blob/main/benchmarks/pypsa/README.md) and automatically generate the LP files ready for benchmarking.**


I encountered an issue during CI with the following error:

`System.IO.IOException: No space left on device : '/home/runner/runners/2.319.1/_diag/Worker_20240923-082302-utc.log'`

This occurs when running the following command in the GitHub Action:

[...]

PR #48, opened 2024-11-04
[benchmarks/pypsa] Generate LP files automatically using Dockerfile

siddharth-krishna:
For now, I've not added the `generate.sh` step to the Dockerfile, so that it can be run manually and debugged if necessary. That way also we can reuse a container if we make small modifications to the benchmarks or scripts.

Partial progress towards #14


jacek-oet (approved):
LGTM!

PR #54, opened 2024-11-07
[fix] Improved test data matches metadata and some housekeeping

jacek-oet:
- Moved the assertion to check that the set of benchmark names in the metadata matches those in the data from home.py to app.py.
- Added a load\_metadata function for reusability and better code organization.
- Updated the Compare page to handle cases where the number of solvers is 1 or 0.

### DIFF
--- a/website/app.py
+++ b/website/app.py
@@ -1,4 +1,10 @@
+from pathlib import Path
+
+import pandas as pd
 import streamlit as st
+
+# local
+from utils.file_utils import load_metadata
 
 st.markdown(
     """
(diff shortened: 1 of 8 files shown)


website/app.py
```diff
@@ -41,5 +47,26 @@
...
+metadata_df = pd.DataFrame(metadata).T.reset_index()
+metadata_df.rename(columns={"index": "Benchmark Name"}, inplace=True)
+# Load the data from the CSV file
+data_url = Path(__file__).parent.parent / "results/benchmark_results.csv"
+raw_df = pd.read_csv(data_url)
+df = raw_df
+
+
+# Assert that the set of benchmark names in the metadata matches those in the data
+csv_benchmarks = set(raw_df["Benchmark"].unique())
```
siddharth-krishna, line 62:
```suggestion
data_df = pd.read_csv(data_url)


# Assert that the set of benchmark names in the metadata matches those in the data
csv_benchmarks = set(data_df["Benchmark"].unique())
```

website/compare.py
```diff
@@ -41,23 +34,36 @@ def load_metadata(file_path):
...
+# Set default solvers based on available solver list length
+if len(solver_list) > 1:
+    default_solver1 = solver_list[0]
+    default_solver2 = solver_list[1]
+elif len(solver_list) == 1:
+    default_solver1 = solver_list[0]
+    default_solver2 = None
+else:
+    st.write("No solvers available for comparison.")
+    st.stop()
```
siddharth-krishna, line 48:
```suggestion
else:
    st.write("Fewer than 2 solvers available for comparison. Please add more solvers to compare.")
    st.stop()
```
We can also simplify the code below to assume that there are at least 2 solvers.

siddharth-krishna (approved):
Looks good, thanks! I left some suggestions to simplify the code.

Can we also quickly add something like `streamlit run` to the CI? Just to check that the website loads without errors. Streamlit must have some method to test this without running with a timeout or something. No need to spend too much time on it if there isn't something out-of-the-box. Thanks

jacek-oet:
@siddharth-krishna 
> Can we also quickly add something like streamlit run to the CI? 

Let me research and implement it in new PR.

website/compare.py
```diff
@@ -41,23 +34,35 @@ def load_metadata(file_path):
...
+# Dropdown to select Solver 2 only if there’s more than one solver
+if default_solver2:
+    solver2 = st.selectbox(
+        "Select Solver 2",
+        solver_list,
+        index=solver_list.tolist().index(default_solver2),
+    )
+else:
+    st.write("Only one solver available. Please add more solvers to compare.")
+    st.stop()
```
siddharth-krishna, line 65: This check and the else block are unnecessary since we check that there are at least 2 solvers above.

siddharth-krishna (approved):

PR #55, opened 2024-11-11
Runner: run multiple versions of each solver 

jacek-oet:


runner/run_benchmarks.py
```diff
@@ -4,12 +4,27 @@
...
+        if solver_name == "highs":
+            return version("highspy")
+        elif solver_name == "scip":
+            return version("PySCIPOpt")
+        elif solver_name == "gurobi":
+            return version("gurobipy")
+        else:
+            return version("glpk")
+    except PackageNotFoundError:
+        return ValueError(f"Package Not Found Error {solver_name}")
```
siddharth-krishna, line 25:
```suggestion
        elif solver_name == "glpk":
            return version("glpk")
        else:
            raise NotImplementedError(f"We do not yet support {solver_name}")
```

Again, it's better to explicitly raise an error like in my suggestion so that when we add a new solver we get a meaningful error. Also note that you `raise` an error, not `return` it.

runner/run_benchmarks.py
```diff
@@ -37,9 +52,13 @@ def parse_memory(output):
     raise ValueError(f"Could not find memory usage in subprocess output:\n{output}")
 
 
-def write_csv_headers(results_csv, mean_stddev_csv):
+def write_csv_headers(results_csv, mean_stddev_csv, override):
+    if override:
+        mode = "w"
+    else:
+        mode = "a"
```
siddharth-krishna, line 59:
```suggestion
    mode = "w" if override else "a"
```

runner/run_benchmarks.py
```diff
@@ -235,17 +273,20 @@ def main(benchmark_file_path, solvers, iterations=1, timeout=15 * 60):
 
 
 if __name__ == "__main__":
-    # Check for benchmark file argument
+    # Check for benchmark file argument and optional year and override arguments
     if len(sys.argv) < 2:
-        print("Usage: python run_benchmarks.py <path_to_benchmarks.yaml>")
+        raise ValueError(
+            "Usage: python run_benchmarks.py <path_to_benchmarks.yaml> [<year>] [<override>]"
+        )
```
siddharth-krishna, line 280: It's getting complex enough that perhaps in the next PR we should switch to using the `argparse` library?

runner/requirements-2020.txt
```diff
@@ -0,0 +1,7 @@
+glpk==0.4.7
+gurobipy==9.5.0
+highspy==1.5.0.dev0
```
siddharth-krishna, line 3: If a solver's first release was in e.g. 2022, then let's not include it in `requirements-2020.txt` and let's change `run_benchmarks.py` to only run the solvers that are present in the pip environment. You can query `linopy.available_solvers` to find out which ones are available.

siddharth-krishna (commented):
Thank Jacek, it looks good overall, but I left some suggestions and requests. Also, let's make a separate PR to add a CI test for streamlit, especially if you switch from what you have here to something like https://docs.streamlit.io/develop/concepts/app-testing/automate-tests#running-the-app-tests

results/benchmark_results.csv
```diff
@@ -1,19 +1,21 @@
...
-pypsa-eur-elec-op-ucconv-3-24h,glpk,TO,Timeout,,900,182.904,0,0
-pypsa-eur-elec-op-ucconv-3-24h,scip,TO,Timeout,,900,760.456,0,0
-pypsa-gas+wind+sol+ely-1-1h,highs,ok,optimal,31714138111.861046,10.052644968032837,464.252,,
-pypsa-gas+wind+sol+ely-1-1h,glpk,ok,optimal,31714138110.0,137.99040961265564,299.336,,
-pypsa-gas+wind+sol+ely-1-1h,scip,ok,optimal,31714138111.860943,173.0952386856079,755.6,,
-pypsa-gas+wind+sol+ely-ucgas-1-1h,highs,ok,optimal,84667526618.3079,104.27388262748718,526.808,0,0
-pypsa-gas+wind+sol+ely-ucgas-1-1h,glpk,ok,optimal,84667526620.0,307.3840732574463,306.436,0,0
-pypsa-gas+wind+sol+ely-ucgas-1-1h,scip,ok,optimal,84667526618.30772,429.5963785648346,852.612,0,0
+Benchmark,Solver,Status,Termination Condition,Objective Value,Runtime (s),Memory Usage (MB),Max Integrality Violation,Duality Gap,Solver Version,Solver Release Year
+pypsa-eur-sec-2-24h,highs,ok,optimal,15337522234.402815,52.85370707511902,431.176,,,,2024
```
siddharth-krishna, line 2: Why is there no value for `Solver Version` for most lines?

runner/run_benchmarks.py
```diff
@@ -37,9 +51,10 @@ def parse_memory(output):
     raise ValueError(f"Could not find memory usage in subprocess output:\n{output}")
 
 
-def write_csv_headers(results_csv, mean_stddev_csv):
+def write_csv_headers(results_csv, mean_stddev_csv, override):
+    mode = "w" if override else "a"
```
siddharth-krishna, line 55: Why would I ever need to append CSV headers onto an existing file? You already have an if statement at the place where this function is called that only calls this function if overriding/writing a new file. So can't the mode in this function always be `w`?

runner/requirements-2022.txt
```diff
@@ -0,0 +1,4 @@
+highspy==1.5.0.dev0
+linopy==0.3.14
+pypsa==0.31.1
```
siddharth-krishna, line 3: Do we need `pypsa` for the runner? 🤔

siddharth-krishna (changes requested):
Thanks, Jacek. A couple of more small things to be done before we can merge this. 🙏 

>  All pages of the website other than the Solver History page should only use the data for the latest version of each solver

Did you do this already? If not, please implement that as well, thanks.

runner/run_benchmarks.py
```diff
@@ -196,13 +251,28 @@ def main(benchmark_file_path, solvers, iterations=1, timeout=15 * 60):
             raise ValueError("No valid 'path' or 'url' found for benchmark entry.")
 
         for solver in solvers:
+            try:
+                solver_version = get_solver_version(solver)
```
siddharth-krishna, line 255: Right now this is being called once for every benchmark. That could be slow if `conda list` takes some time. Can we make a `solver_versions` dict that maps each solver in `solvers` to its version before we start the `for file_info in benchmark_files_info:` loop? That way we only query each solver's version once.

siddharth-krishna (commented):
Thanks, Jacek! I left 2 small suggestions, but this looks good otherwise. I presume you've tested it locally on the 2 sample LP files or the smallest benchmarks?

If you could quickly implement these suggestions, then I can run all the benchmarks on all the solver versions and update the results file in this PR before we merge it in. 🙏

runner/benchmark_all.sh
```diff
@@ -0,0 +1,62 @@
...
+
+    env_name="env-$year"
+
+    conda create -n "$env_name" python=3.9 -y
+    source "$(conda info --base)/etc/profile.d/conda.sh"  # Ensure conda is initialized
+    conda activate "$env_name"
+
+    if [ -f "$req_file" ]; then
+        pip install --upgrade pip
+        pip install -r "$req_file"
```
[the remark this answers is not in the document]
jacek-oet: As we have removed the requirements files, it is necessary to update the CI's 'Run sample benchmarks' step. I will create a separate PR to address this update.

.github/workflows/ci.yaml
```diff
@@ -39,20 +38,7 @@ jobs:
       - name: Install Dependencies for Streamlit App Action
         run: |
             pip install -r website/requirements.txt
-
       - name: Run Streamlit App Action
         uses: streamlit/streamlit-app-action@v0.0.3
         with:
             app-path: website/app.py
-
-      - name: Run sample benchmarks
```
siddharth-krishna, line 48: Why was this CI test removed?

siddharth-krishna (approved):

.github/workflows/ci.yaml
jacek-oet: @siddharth-krishna I commented about that here.
https://github.com/open-energy-transition/solver-benchmark/pull/55#discussion_r1850388500

jacek-oet:
@siddharth-krishna LGTM! (As author of this PR, I am not permission to approve.)

PR #59, opened 2024-11-19
Website: add SGM to home dashboard

jacek-oet:


siddharth-krishna (approved):
Thanks, Jacek, this is perfect. I pushed a tiny edit, to sort solvers by the least SGM so the best one is on top. If you agree, please merge this in!

PR #61, opened 2024-11-20
Add multiple sizes of each benchmark and analyze performance scaling

siddharth-krishna:
This PR makes the following changes:
- `generate-sizes.sh`: Generates multiple sizes of the pypsa-eur benchmarks by varying the spatial and temporal resolutions. This script replaces `solver_benchmark_pypsa_eur.py` and `generate.sh`, I'll update the README shortly
- Modifies the format of `metadata.yaml` to have a list of sizes of each benchmark, and move the size-dependent fields like "number of variables" into each size of the benchmark.
- After generating 36 sizes of the pypsa-eur benchmarks, the `filter-benchmarks.{sh,py}` scripts run Highs on them to filter out the ones that can be solved in under 1h
- The generated LP files are now being saved to GCP buckets instead of GDrive as this can be automated using `gsutil -m cp runner/benchmarks/*.lp <GCP-bucket-path>`
- The `benchmark_config.yaml` format has also been updated to have a list of sizes for each benchmark
- The runner has been modified to read the new format of `benchmark_config.yaml`, download LP files from the GCP bucket, and also save the size information of each benchmark to the result CSV files
- And thanks to Jacek:
  - The website has been updated to work with the new size data in the result CSV files
  - A new performance scaling page has been added to the website to compare performance vs benchmark size

To do on the generation/runners side:
- [x] Adapt the generation script to work on all benchmarks
- [x] Automatically query generated LPs for num vars etc and put into metadata file ~(also get num int/continuous vars) -- Daniele is adding this in the next PR!~
- [x] Adapt runner to work on sizes and write size info to results CSV

To do on the website side:
- [x] Adapt `home.py`, `compare.py`, and `history.py` to work with the new CSV format, but displaying all the data for all the sizes
- [x] `benchmarks.py` should extract the information under `Sizes:` in `metadata.csv` and display it in a separate table, showing the sizes available for each benchmark
- [x] Add a new page `scaling.py` that plots {runtime, memory consumption} vs {spatial resolution, num variables, num constraints} as in [this](https://discord.com/channels/1097854148965306470/1308340506824278047/1308340517024694323) picture, but with one line plot per benchmark. Also implement Daniele's suggested improvements to the pic. Here is the code I used as a reference:

```python
figures = [
    px.scatter(data, x='Spatial Resolution', y='Runtime (s)', color='Solver'),
    px.scatter(data, x='Num Variables', y='Runtime (s)', color='Solver'),
[...]

PR #64, opened 2024-11-22
Add jump-highs platform benchmarks

danielelerede-oet:
Hi @siddharth-krishna , I added the metadata for all the GenX, PowerModels and Tulipa benchmarks on the JuMP-Highs platform. Some notes:

- The Sienna ones are still missing (I hope to be able to work on them on Monday)
- For the first time I added continuous variables as size parameters (I was thinking to do so for PyPSA benchmarks, too)
- Tulipa benchmarks are also classified by size as in the current version of PyPSA metadata
- We could work on increasing the number of GenX-based benchmarks removing unit commitment from at least some of the benchmarks on the platform (as they're all MILP at the current status)


benchmarks/jump_highs_platform/Metadata_tulipa.yaml
```diff
@@ -0,0 +1,42 @@
...
+  Short description: European-level investment and operation model to consider integer investments and unit commitment variables
+  Model name: Tulipa
+  Version:
+  Technique: MILP
+  Kind of problem: Infrastructure
+  Sectors: Power
+  Time horizon: Single period (1 year)
+  MILP features: Unit commitment
+
+Sizes:
```
siddharth-krishna, line 12: @danielelerede-oet this entire block should be indented so that it is under the `1_EU_investment_simple:` key above. And we don't need names for each benchmark size, those are auto-generated from the parent key e.g. `1_EU_investment_simple` and the temporal & spatial resolutions. 

Perhaps we could also put the URLs as a separate line, so that it doesn't mess with the code that extracts the benchmark name? This change needs to be done in the other files as well. 🙏 

Here's an example:

```yaml
pypsa-eur-sec:
  Short description: Sector-coupled PyPSA-Eur infrastructure run for Italy considering 2050 as single planning horizon (LP, lot of variables, strongly intermeshed constraints)
  Model name: PyPSA-Eur
  Version: 0.13.0 (commit 8f1a6b1)
  Technique: LP
  Kind of problem: Infrastructure
  Sectors: Sector-coupled (power + heating, biomass, industry, transport)
  Time horizon: Single period (1 year)
  MILP features: None
  Sizes:
  - Temporal resolution: 24
    Spatial resolution: 2
    N. of constraints: 132272
    N. of variables: 63264
    URL: https://github.com/jump-dev/open-energy-modeling-benchmarks/blob/main/instances/TulipaEnergyModel_1_EU_investment_simple_24h-dc195d31e02ab753b5f047a4e754d2d387b59e15ac813c515b7fab21651779ea.mps.gz
  - Temporal resolution: 24
    Spatial resolution: 3
    N. of constraints: 132272
    N. of variables: 63264
    URL: https://github.com/jump-dev/open-energy-modeling-benchmarks/blob/main/instances/TulipaEnergyModel_1_EU_investment_simple_24h-dc195d31e02ab753b5f047a4e754d2d387b59e15ac813c515b7fab21651779ea.mps.gz
```

benchmarks/jump_highs_platform/Metadata_genx.yaml
```diff
@@ -0,0 +1,140 @@
...
+  Version:
+  Technique: MILP
+  Kind of problem: Infrastructure
+  Sectors: Power
+  Time horizon: Single period (1 year)
+  Temporal resolution: Hourly
+  Spatial resolution: 3 nodes (Massachussetts, Connecticut, Maine)
+  MILP features: Unit commitment
+  N. of constraints: 171878
+  N. of variables: 171880 (continuous), 16635 (integer), 3696 (binary)
```
siddharth-krishna, line 14: Could we please split these into 3 lines, so that each value is an integer? That way the code in #61 can parse it automatically to generate the performance scaling plots.

Btw, could you share the code you use to calculate these 3 features? Eventually we'll make a script to extract this data automatically. 🚀

siddharth-krishna (commented):
Thanks, Daniele! This gives us a good starting point to work on the benchmark runner and get some results already. I think perhaps we can save the task of making LP versions of GenX benchmarks for after the 2nd quarter check-in if you don't have time.

@jacek-oet could you perhaps help with the following tasks on this branch to add these benchmarks to the runner and website? 🙏 

- [ ] Add the names and URLs from the metadata files added by Daniele to `benchmarks/benchmark_config.yaml` so that the runner downloads them and runs them
- [x] Modify the runner to unzip downloaded files that end in `.gz` before running
- [x] Check that the runner can run MPS files by running all these benchmarks with a super short timeout. I tested locally that the existing code in `run_solver.py` works with MPS files, so it should work.
- [ ] I'm not sure what the best way to integrate the benchmark metadata into the website, but here's a proposal (please do suggest a better alternative if you see one!): make a script that parses all the `metadata*.yaml` files under `benchmarks/` and creates a unified `results/metadata.yaml` file. The website should use this file instead of `benchmarks/pypsa/metadata.yaml`, as it contains info for all benchmarks not just PyPSA.
- [ ] Test that all pages in the website shows the new benchmarks, and work as intended.

danielelerede-oet:
Hi @siddharth-krishna , I added metadata for Sienna benchmarks. Unfortunately, differently for the other modeling platforms, they don't provide a full access to data, so I had to dig out information as far as I could, but as you can see there are a lot of details that I don't have access to... For this, the only choice is to ask the jump-highs people

danielelerede-oet:
Hi @siddharth-krishna , I also added the script to read information about size of the benchmarks from .mps files!

siddharth-krishna (approved):
Thanks, Jacek, this looks good. Just a minor comment, but it can also be done in another PR next week if you don't have time today 🙏

siddharth-krishna:
I did a test run of all the benchmarks so far on all the solver versions we have that work (total 959 combinations!), and just pushed the results to this branch. Looks like overall, our benchmarking infrastructure is working, great job everyone 🚀 

I had to make the following changes in order to do the run, please can you review when you have time:
- Update `benchmark_config.yaml` to the new structure including sizes
- Added Sienna benchmarks to `benchmark_config.yaml`
- Renamed some Jump-Highs benchmarks so that they begin with the model framework name
- Fix merge_metadata.py script to output metadata in the new sizes format (please note, @jacek-oet ), and re-ran it to regenerate `results/metadata.yaml`
- Runner: update to support MPS files

If all this looks good, I think we can merge this PR in, right?

jacek-oet (approved):
LGTM

PR #71, opened 2024-12-04
Website: Fix missing lib & column

jacek-oet:


siddharth-krishna (approved):
Thanks! Really appreciate the super quick fix 🙌

PR #72, opened 2024-12-04
Correct Tulippa metadata and do a 1 min run on C4

danielelerede-oet:
Hi @siddharth-krishna, here I correct metdata for Tulipa according to the mistake we noticed from the website + I add "units of measurement" for spatial and temporal resolution in the metadata for other modelling platforms + I change the order in which "N. of variables" and "N. of constraints" are presented in the merged metadata file


siddharth-krishna:
Thanks, Daniele! Unfortunately, we had intentionally removed the units such as `nodes` so that the website could plot the runtime vs spatial/temporal resolution in the Performance Scaling page.. My thought was that these units could be added back in the frontend website code when displaying the metadata?

But perhaps it's also useful to have it in the raw metadata YAML files? If so, alternatively, we could change the website code that parses the merged metadata file to strip the units and then plot the graphs.

Also, I'm curious why we prefer to order constraints before variables? :)

danielelerede-oet:
Hi @siddharth-krishna ok, makes sense. Actually yes, it would be better to have units in metadata (Jonas also raised this point yesterday, e.g., highlighting that it's useful to have a clear indication about how spatial resolution is taken into account) but if it's easier for the moment let's just keep it as it is.
Concerning the order of constraints and variables, it was just to avoid having something like "n. of variables", "n. of constraints" and then again "n. of continuous variables" etc.

siddharth-krishna:
Thanks, Daniele. Both your points make sense, but given our Tuesday deadline, how about we just fix the temporal resolution for the Tulipa benchmarks and the order of constraints/variables in this PR and keep the remaining fixes regarding units for after Tuesday? If that sounds reasonable, I can push a commit to this branch doing that. 🙏 

(Another little hitch is that we use the values in spatial/temporal resolution fields of metadata to programmatically generate filenames, so having units or something like `N/A` (as opposed to `NA`) will result in potentially broken filenames e.g. `pypsa-eur-sec-10 nodes-N/A.lp`, so we need to fix the code in a few more places to support that.)

danielelerede-oet:
> Thanks, Daniele. Both your points make sense, but given our Tuesday deadline, how about we just fix the temporal resolution for the Tulipa benchmarks and the order of constraints/variables in this PR and keep the remaining fixes regarding units for after Tuesday? If that sounds reasonable, I can push a commit to this branch doing that. 🙏
> 
> (Another little hitch is that we use the values in spatial/temporal resolution fields of metadata to programmatically generate filenames, so having units or something like `N/A` (as opposed to `NA`) will result in potentially broken filenames e.g. `pypsa-eur-sec-10 nodes-N/A.lp`, so we need to fix the code in a few more places to support that.)

Yes, sounds totally reasonable. Ok, thank you for the explanation on `NA`, if that works better for you it's definitely not a problem for me! :)

siddharth-krishna (approved):
@danielelerede-oet @jacek-oet I added to this PR the results of running all the benchmarks with a 1 min TO on a GCP C4 VM. I'm merging this in now so that we can show these results in today's presentation, but please do review my commits when you have some time and let me know if I can make any changes/improvements. 🙏

PR #77, opened 2024-12-10
[website] Rename the size column to instance, fix size filters

jacek-oet:
image

image
image
image
image
image


siddharth-krishna (approved):
Thanks, Jacek. One more quick request: can you add the units e.g. `3 nodes` and `24h` to the values in the Sizes table in the Benchmark details page? This is not urgent, please fix the size filter first. Thanks!

siddharth-krishna (approved):

PR #79, opened 2024-12-10
Website: Add Min and Max Runtime to Results table

jacek-oet:


website/home.py
```diff
@@ -113,13 +113,19 @@ def combine_sgm_tables(df):
         runtime_values = group["Runtime (s)"]
         sgm_runtime = calculate_sgm(runtime_values)
 
+        # Calculate Min and Max Runtime
+        min_runtime = str(round(runtime_values.min(), 1))
+        max_runtime = str(round(runtime_values.max(), 1))
```
siddharth-krishna, line 118: Do we need to convert to `str`? It changes the alignment in the final table, it's better if it is a float so it is right aligned.

siddharth-krishna (commented):
Thanks, Jacek. Is it possible to remove the TO values when calculating the max runtime? Right now it's always equal to `60`s which is the TO

siddharth-krishna (approved):
Thanks!!

Issue #68, opened 2024-12-02
GLPK errors on JuMP-HiGHS MPS benchmarks


siddharth-krishna:
@jacek-oet @danielelerede-oet I notice in the results on https://github.com/open-energy-transition/solver-benchmark/pull/64 (commit [59f1e54](https://github.com/open-energy-transition/solver-benchmark/pull/64/commits/59f1e545fcadbdecd14d039ae3c1dc7e2d29149f)) that GLPK errors on all the newly added MPS benchmarks, so I ran one manually to see the problem:

```
glpsol --mps runner/benchmarks/genx-7_three_zones_w_colocated_VRE_storage-3-24h.mps 
GLPSOL--GLPK LP/MIP Solver 5.0
Parameter(s) specified in the command line:
 --mps runner/benchmarks/genx-7_three_zones_w_colocated_VRE_storage-3-24h.mps
Reading problem data from 'runner/benchmarks/genx-7_three_zones_w_colocated_VRE_storage-3-24h.mps'...
runner/benchmarks/genx-7_three_zones_w_colocated_VRE_storage-3-24h.mps:1: warning: missing model name in field 3
Objective: Obj
runner/benchmarks/genx-7_three_zones_w_colocated_VRE_storage-3-24h.mps:19573: in fixed MPS format positions 37-39 must be blank
MPS file processing error
```

I can't see anything obviously wrong in the MPS file on that line. Googling for the error message leads to this issue:
https://github.com/Pyomo/pyomo/issues/2115

But I can't understand the difference between the 2 MPS file in the OP, it looks like there are just white space differences? Any ideas what the problem could be? (cc @FabianHofmann in case you've seen this error before)
![image](https://github.com/user-attachments/assets/677a3f24-3bbd-4e47-880e-31fc54238491)

Another problem is that currently we are recording such errors like so:
```csv
Benchmark,Size,Solver,Solver Version,Solver Release Year,Status,Termination Condition,Runtime (s),Memory Usage (MB),Objective Value,Max Integrality Violation,Duality Gap
genx-1_three_zones,3-1h,glpk,5.0,2020,warning,unknown,0.3038334846496582,170.26,nan,,
```
[...]

PR #86, opened 2024-12-20
Re-run GLPK results after fixing linopy MPS format

siddharth-krishna:
The errors we were seeing with GLPK in #68 was because linopy by default calls `glpsol` with the `--mps` argument, which is for the fixed MPS format, but the JuMP-HiGHS benchmarks are using the free MPS format (`--freemps`), so I re-ran the GLPK runs by hacking linopy to use `--freemps`:
```diff
diff --git a/linopy/solvers.py b/linopy/solvers.py
index e89a6a9..2c996e3 100644
--- a/linopy/solvers.py
+++ b/linopy/solvers.py
@@ -563,7 +563,8 @@ class GLPK(Solver):
         Path(solution_fn).parent.mkdir(exist_ok=True)
 
         # TODO use --nopresol argument for non-optimal solution output
-        command = f"glpsol --{io_api} {problem_fn} --output {solution_fn} "
+        io_api_arg = "freemps" if io_api == "mps" else io_api
+        command = f"glpsol --{io_api_arg} {problem_fn} --output {solution_fn} "
         if log_fn is not None:
             command += f"--log {log_fn} "
         if warmstart_fn:
```

This PR has the new results, and I'll open a linopy PR to fix the issue in a more robust way in the new year. (FYI @FabianHofmann )

There are still a few benchmarks returning `uknown` statuses, this needs to be looked into.


danielelerede-oet (approved):
Hi Sid! Looks great, any hint about the reason why we get the `unknown` status?

siddharth-krishna:
Thanks, Daniele. In both cases, I see an error similar to the following. I suppose we can take this as a bug/error in GLPK, and leave it as is? HiGHS and SCIP are able to solve one of them successfully.
```
Solving LP relaxation...
GLPK Simplex Optimizer 5.0
5664 rows, 7218 columns, 25566 non-zeros
      0: obj =   9.127091686e+04 inf =   2.551e+02 (73)
Error: unable to factorize the basis matrix (1)
glp_intopt: cannot solve LP relaxation
```

PR #87, opened 2025-01-07
Benchmarks: new GenX "real" benchmarks

danielelerede-oet:
Hi @siddharth-krishna , I added metadata and config files for the GenX extended benchmarks built starting from Gabe's sample problem. I tried to combine all the available features, developed benchmarks using the pre-defined temporal resolution in Gabe's folder (weekly/168 h) and got .lp files for all of them. 
I identified 3 main "issues":
- I tried to develop benchmarks with different temporal sizes (e.g. monthly/730 h) but they take forever to build the model
- I also tried to solve the benchmarks but any attempt gets killed due to memory issue
- Indeed, the generated .lp files are huge (ranging from 1.1 to 2 GB) and you can see from metadata how the size of the problems is definitely large.
Now, the .lp files are definitely too large to be shared here, so I'm uploading them on GDrive https://drive.google.com/drive/folders/1T6jVUJEFSMQv2xzMUS5f18OIArDOGzQ1?usp=drive_link .

To run the GenX extended benchmarks (after GenX installation - using Julia 1.9, the latest version is not compatible with GenX) I adopted the following workflow:
-  I cloned the repo at https://github.com/GenXProject/GenX.jl
- I copied the `case_for_oet` folder provided by Gabe in GenX.jl
- I substituted the `GenX.jl/case_for_oet/settings/genx_settings.yml` with any of the .yml files pushed here and changing their name to `genx_settings.yml` (each config file already contain a line to print out the .lp files)
- I run each case in the GenX.jl Julia project using `include("/home/user/solver-benchmark/benchmarks/genx/GenX.jl/case_for_oet/Run.jl")`
This workflow takes advantage of the content of the `GenX.jl/case_for_oet/TDR_Results` folder for inputs clustered according to a time resolution of 168 h (as reported in `GenX.jl/case_for_oet/settings/time_domain_reduction_settings.yml`)

Whenever the temporal resolution of the model requires changes one should adopt a simple workflow:
- if already present, the `TDR_Results` folder should be deleted (or its name should be modified)
- the `TimestepsPerRepPeriod` entry in `GenX.jl/case_for_oet/settings/time_domain_reduction_settings.yml` should be adjusted according to the desired temporal resolution (e.g. 1 for hourly resolution, 24 for daily resolution, 168 for weekly resolution, 730 for monthly resolution - I noticed that any different time resolution than the ones indicated here causes failure of the model run)
- The model can be run using again `include("/home/user/solver-benchmark/benchmarks/genx/GenX.jl/case_for_oet/Run.jl")` and a new `TDR_Results` folder is generated according to the new temporal resolution


siddharth-krishna:
Thanks, Daniele, this is great work. Am I correct in assuming that #40 was closed because its benchmarks are anyway present in the JuMP-HiGHS set and in this PR?

Is the intention with this PR to add "real" sized benchmarks only? If so, we can try solving the LP files you have uploaded with Gurobi on a larger VM/ZIB to check that it solves in <10h.

If we are also planning to get small & medium sized benchmarks from this PR, is it possible to tweak the configs to get smaller LPs that you can solve locally? Or is the plan to get S & M benchmarks from another source?

danielelerede-oet:
Hi @siddharth-krishna , the benchmarks in #40 were originated from the problems in the JuMP-Highs platform, but don't add anything special to them, so I removed them in favor of these extended (in terms of spatial scope mainly, since it's difficult to scale up GenX benchmarks in size) benchmarks.

With this PR we are adding only real-size problems, exactly, and I would definitely go with testing the .lp files I got on a more powerful VM. S & M benchmarks cannot definitely be obtained starting from Gabe's sample problem, but I think we already have good S & M samples in the problems from the JuMP-Highs platform

danielelerede-oet:
@siddharth-krishna instructions added for GenX, too.

siddharth-krishna (approved):

PR #91, opened 2025-02-03
Benchmarks: new PyPSA benchmarks with missing features

danielelerede-oet:
Hi @siddharth-krishna I added 3 new PyPSA benchmarks to cover the missing features in the single node capacity expansion small benchmarks. So we have a new small problem with the addition of a battery storage for electricity, one with an emission limit and one with battery storage and unit commitment on the gas generator (so 2 LPs and 1 MILP). I also wanted to add CCS and retrofits among the technological options but the current technology-data repo does not contain data to include them in the power model. It may be considered as a possible future update.
I also modified the `only_generate_problem_file` option in `keep_files` to get .lp files as the latest linopy version doesn't allow its use anymore (I actually don't remember if you used a customized solution for not solving the problem but just getting the .lp file, but I checked the possible options for the `n.optimize` function and `only_generate_prolem_file` is not there).
I didn't add the size statistics as we were mentioning about using a JuMP-Highs tool for getting them.


siddharth-krishna (commented):
Thanks, Daniele. As discussed, I'm moving the changes to benchmark_config.yaml from #93  to here.

siddharth-krishna (approved):

siddharth-krishna (approved):

PR #93, opened 2025-02-06
update benchmark_config

danielelerede-oet:
Hi @siddharth-krishna here is the updated `benchmark_config.yaml` with all the new benchmark files. I used the public URL on GCP. Also, the files `pypsa-gas+wind+sol+ely-1-1h.lp` and `pypsa-gas+wind+sol+ely-1-1h-ucgas.lp` can be removed from there.
Concerning temoa benchmarks, sizes are not reported with the format `x-xh` because temoa doesn't work with hourly resolutions. Hope that's not an issue, otherwise we can use a temporary placeholder for that.


siddharth-krishna:
Thanks, Daniele! The temoa size naming shouldn't be an issue, but I will test it with the benchmark runner script now to check.

If you don't mind, I will take the changes from this PR and apply them to the respective PRs that introduce the new benchmarks, so that I can test the changes in each PR individually.

siddharth-krishna:
Closing this PR as changes have been applied to #87, #89, and #91

Issue #83, opened 2024-12-13
Build a production website


siddharth-krishna:
Let's build a final production-quality website using modern technologies and the professional design from Heba's Figma prototype. Here is @jacek-oet 's proposed stack:
- Web framework: NextJS
- Style: Tailwind CSS
- Chart: D3.js
- API Layer: RESTful APIs

Here's the detailed plan, each item can be converted into an issue to track the subtasks:

- [x] Setup source (Chart, Tailwind, Font, Color, icon)
- [x] Landing Page
- [x] Home Page
  - [x] Get num solvers, benchmarks, TO, etc from results
  - [x] Filters: also show "only" on hover to allow filtering to only one thing
  - [x] Results table: only show the `v` on column that is sorted, allow sorting by columns on clicking column header
  - [x] Benchmarks plot: also show benchmark name and size on hover
  - [x] Can we change the top bar to also show versions and sizes? E.g. Solvers: 3 (7 versions), Benchmarks: 58 (117 instances)
  - [x] Change the header Results to Results (filtered to X/Y benchmarks) when filters are active
  - [x] Have a ? icon next to Solved Benchmarks which shows the tooltip "Solved benchmarks is the number of benchmarks where the solver returns an ok status"
  - [x] Remove the Benchmarks header, and move the next line of text x represents ... to below the graph
  - [x] Column headers: Runtime -> SGM Runtime, Memory -> SGM Memory
  - [x] Add help icon explaining SGM (https://plato.asu.edu/ftp/shgeom.html)
  - [x] Solved benchmarks column: show 23 / 60 or whatever,
  - [x] Results table: default sort is by SGM Runtime
  - [x] bug in the calculation of the numbers in the header "Results(filtered to 381/411 benchmarks)"
- [x] Benchmarks details Page
  - [x] Main page listing all benchmarks with filters
  - [x] Details of a single benchmark page
    - [x] Typo: "No. Constrains" should be "No. Constraints"
    - [x] Back button to go back to all benchmarks
    - [x] Have a smaller max height for Instances and MILP features tables
    - [x] Are we missing the MIP Information table? E.g. see `pypsa-eur-elec-op-ucconv`
- [x] History details Page
- [x] Compare solvers Page
    - [x] Change background color of plots to match colors of dropdown, see Figma
- [x] Raw results data Page
  - [x] Can we show all the columns that are shown in https://solver-benchmark.streamlit.app/raw-results ?
[...]

PR #100, opened 2025-02-14
Website: some improvements and fixes

jacek-oet:
Completes #83


siddharth-krishna (commented):
Thanks so much, Jacek, these are great improvements to the website! I would like to request please the following fixes as soon as possible:

- [x] Use percentages in the Solved Benchmarks column, e.g. 08% (6 / 77), so that string based sorting does the right thing. Otherwise right now if you filter to MILPs and sort you get the wrong order:
image
- [ ] The number of benchmarks and instances in the top bar should not depend on the filters. It's meant to be the overall summary of our benchmark set. Also, why is it always showing 3? I think it should be a higher number, around 58...
image
image
- [ ] Can we add scroll bars to the filters so that it doesn't show more than 2 lines? Right now it looks a bit too big..
- [ ] Please correct the background colors in these graphs: the upper triangle should be the same background color as the Solver 1 dropdown and the lower traingle should be the same background color as the Solver 2 dropdown. And both graphs should have the same background colors.
image

- [ ] The values in the filter (e.g. TEMOA, or Sector coupled) should not be hard coded. Please can you get all available options from the metadata file? Otherwise it'll be hard to update the filters as we add new benchmarks.

website-nextjs/src/components/admin/ResultsSections.tsx
```diff
@@ -206,6 +206,18 @@ const ResultsSection = () => {
     latestBenchmarkResult.map((result) => `${result.benchmark}-${result.size}`)
   ).size
 
+  const getSolvedBenchmarksLabel = (
+    solver: string,
+    uniqueBenchmarkCount: number
+  ) => {
+    const numberSolvedBenchmark = getNumberSolvedBenchmark(solver)
+    const percentage = numberSolvedBenchmark / uniqueBenchmarkCount
```
siddharth-krishna, line 214:
```suggestion
    const percentage = numberSolvedBenchmark * 100 / uniqueBenchmarkCount
```
:)

siddharth-krishna (approved):
Thanks, I've moved remaining tasks to #106. Please add this one suggestion, merge main, and merge this PR. 🙏

PR #105, opened 2025-02-17
Benchmarks: correct metadata for some MILPs

danielelerede-oet:
Hi @siddharth-krishna , I made corrections to some PyPSA metadata


siddharth-krishna (approved):
Thanks, Daniele. We should run the `merge_metadata.py` script when you are done to ensure that the updates also propagate to the website. Thanks

danielelerede-oet:
It's all ok now @siddharth-krishna , I've just finished checking

PR #109, opened 2025-02-18
feat: use all available options from the metadata file, quick fixes, …

jacek-oet:
…round number

Fixed:
- DetailSection: change hardcoded timeout from 15min to 10min

- DetailSection: (${avaliableInstance.length} instances) is wrong as it changes when filters are applied, can it always be the total number of instances in the results?

- The values in the filter (e.g. TEMOA, or Sector coupled) should not be hard coded. Please can you get all available options from the metadata file? Otherwise it'll be hard to update the filters as we add new benchmarks.


siddharth-krishna (approved):
Oh wow thanks, this looks good 😎

PR #107, opened 2025-02-17
Add results of CBC solver on e2-standard-2

siddharth-krishna:

PR #110, opened 2025-02-18
Website: Improve Solver Loading and Color Generation

jacek-oet:


siddharth-krishna (approved):
Thanks, this is great, #107 works for me now :)

PR #111, opened 2025-02-19
Website: re-order things in the top details bar

siddharth-krishna:
Re-ordered the items in this top bar:
image

Also, I changed the icon for timeout to be more "time" related. But in the future, can we find something more appropriate, like an hourglass?


jacek-oet (approved):
LGTM

PR #118, opened 2025-02-24
Website: add option to calculate SGM on benchmarks solved by all

### LINKED ISSUE #114: Home: SGM on intersection of benchmarks solved by all
SGM Runtime is currently very misleading on #107 because CBC solves much fewer benchmarks compared to HiGHS:

Image

It would be good to add a toggle/dropdown between "Compute SGM using TO values / penalizing TO by a factor of <> / only on intersection of solved benchmarks" and adding the following tooltips to explain the choices:

When calculating SGM, there are many choices of what to do about benchmarks that error or time-out. "Using TO values" will assign to them the time-out or the maximum value of memory used when running the benchmark instance. "Penalizing TO by a factor of X" will take the TO/max value from the previous option and multiply it by a factor of X. "Intersection of solved benchmarks" will filter the benchmark instances to those that are solved by all solvers before computing SGM, so that there are no error or time-out values to consider.

The code will also need to be updated to compute the SGM appropriately based on what the user selects.

For the "Intersection of solved benchmarks" case, the "SGM Runtime" column header should display the number of benchmarks in the intersection. E.g. "SGM Runtime on 5 benchmarks"

jacek-oet:
closes #114


siddharth-krishna:
Thanks, Jacek! This looks good mostly, except maybe there's a bug when selecting the "intersection of solved benchmarks" option? It says it's calculating SGM Runtime on 141 benchmarks, but there are only 137 instances in total, so how can that be? (Perhaps it's 141/3 = 47 benchmarks?)

image

Also, selecting "intersection of solved benchmarks" brings down the number of benchmarks in the "Solved Benchmarks" column to 64. But that's confusing too -- if we are filtering this column also to the intersection of benchmarks solved by all solvers then I'd expect all the percentages in the solved benchmarks column to be 100%. But the numerators don't change, only the denominators do:
image

Another thing we should do, since the "intersection of solved benchmarks" option is essentially another filter, is that we should also change the header from "Results" to "Results(filtered to X/137 benchmark instances)" with the correct value for X?

@danielelerede-oet I would love to hear your thoughts on this feature too -- is it confusing? Any thoughts on how we can improve this? Thanks!

siddharth-krishna (approved):

PR #124, opened 2025-02-25
Run benchmarks via open tofu

KristijanFaust-OET:


infrastructure/main.tf
```diff
@@ -0,0 +1,113 @@
+terraform {
+    required_providers {
+        google = {
+            source = "hashicorp/google"
+            version = "~> 4.0"
+        }
+    }
+}
+
```
stefan-oet, line 9: I would expect a remote backend here. Does that mean that we currently store the state locally? Best practice here would be to use a gcs bucket with versioning enabled. In case of corrupted state, then you can still revert to previous version.

infrastructure/main.tf
```diff
@@ -0,0 +1,113 @@
...
+    benchmark_file = each.value.filename
+    benchmark_year = lookup(each.value.content, "year", "2024")
+    benchmark_content = file("${path.module}/benchmarks/${each.value.filename}")
+  }
+
+  # Add the startup script from external file
+  metadata_startup_script = file("${path.module}/${var.startup_script_path}")
+
+  service_account {
+    scopes = ["cloud-platform"]
```
stefan-oet, line 103: Maybe there is a more specific scope we could use here? cloud-platform is quite broad

infrastructure/main.tf
```diff
@@ -0,0 +1,113 @@
...
+    default = "startup-script.sh"
+}
+
+locals {
+  benchmark_files = fileset("${path.module}/benchmarks", "*.yaml*")
+  benchmarks = {
+    for file in local.benchmark_files :
+      replace(file, ".yaml", "") => {
+        filename = file
+        content = yamldecode(file("${path.module}/benchmarks/${file}"))
```
stefan-oet, line 64: Do we check somewhere for invalid yamls? Or are they auto generated?

infrastructure/main.tf
```diff
@@ -0,0 +1,113 @@
...
+  # Add the startup script from external file
+  metadata_startup_script = file("${path.module}/${var.startup_script_path}")
+
+  service_account {
+    scopes = ["cloud-platform"]
+  }
+}
+
+# Outputs
+output "instance_ips" {
```
stefan-oet, line 108: Usually outputs and variables are separate tf files but here that's maybe not necessary.

stefan-oet (commented):
Looks good to me! Few small comments from my side. I think we should move the state to a gcs bucket unless there is a good reason not to.

infrastructure/main.tf
KristijanFaust-OET: I'll need to do a follow up PR were I'll introduce GCP buckets for result storing, but also auto destroying the VMs (since every VM will finish benchmarking at a different time). I'm not sure we will rely much on tofu state for machine management yet, but if we do we can consider transferring it to a bucket.

infrastructure/main.tf
KristijanFaust-OET: They are manual for the most part AFAIK. Good idea to check for yaml validity :+1: Will look into it.

infrastructure/main.tf
stefan-oet: I see, for resources with a short lifetime this is probably less crucial at the moment. When the tf code is run from multiple places (ci / other developers) then the remote state becomes more important

infrastructure/main.tf
KristijanFaust-OET: I don't think it is necessary at the moment neither. For now it is just an easy way to quickly obtain the machine ips so that the user doesn't need to go to the web console in order to perform an ssh. I would consider having a separate file if the output grows with time and tofu's state starts to play a role in the process, but as mentioned we probably won't rely much on tofu's state for the runs, if at all.

infrastructure/main.tf
KristijanFaust-OET: True, but the thing is this will probably never be run from multiple places. The idea is to run this only once for a GCP project per benchmarking cycle. There won't really be a need to run it incrementally from multiple places.

infrastructure/main.tf
KristijanFaust-OET: Changed: https://github.com/open-energy-transition/solver-benchmark/pull/124/commits/6f0f9470036a49d54d0583066e21062ffb00d843

I don't really have much experience with this so I'm not sure if this is the best combination, so feel free to suggest some concrete values here if you know some.

infrastructure/main.tf
KristijanFaust-OET: Now tofu should fail if it encounters an error with an error message: https://github.com/open-energy-transition/solver-benchmark/pull/124/commits/24707716b33cd693f132449a9c6eb91e77324fac

infrastructure/main.tf
stefan-oet: looks good to me. maybe the monitoring should also be write instead of read

jacek-oet (approved):
LGTM

jacek-oet (approved):
LGTM

PR #125, opened 2025-02-27
Website: add solvers page & relative performance plot

### LINKED ISSUE #115: Website: add solvers page & relative performance plot from Matthias
It would be nice to add a new dashboard "Solvers" that lets the user pick one of the available solvers (HiGHS, SCIP, CBC, etc.) and shows:
- Some key info about each solver: Name, Website, License, Source Code
- A relative performance bar chart that uses the selected solver as the base solver and shows relative performance factors of all other solvers on all benchmark instances. For example, this is what the plot looks like with base solver HiGHS:
https://mattmilten.github.io/mittelmann-plots/milp_8threads-HiGHS.html
- If it's not hard to do, let's reuse the filters, and only show the filtered benchmark instances in the above plot
- Let's add some text explaining the plots and attributing Matthias as inspiration (I can help with this after you open a PR)

Some suggestions:
- We don't have a design for this page yet, so please just make a quick & easy prototype implementation using existing components/styles and we can design it after we see and test the prototype
- There is some text introducing the plots at https://mattmilten.github.io/mittelmann-plots and the code is available at https://github.com/mattmilten/mittelmann-plots for reference
- No need to implement the "virtual best" solver for now, I'll make a separate issue for that
- Since we're not using plotly we don't need to implement the "click to hide/show" feature for now

jacek-oet:
closes #115 

image


siddharth-krishna:
Thanks, Jacek! I haven't had a chance to look at the code yet, I will do tomorrow, but from first glance of the screenshot I'm wondering if we can make a few improvements:

- [ ] Can we sort the benchmarks in increasing order of runtime of the base solver (glpk in your screenshot)? This is what [Matthias](https://mattmilten.github.io/mittelmann-plots/milp_8threads-HiGHS.html) also does
- [ ] Can we remove the benchmark names from the x-axis labels? I think we have too many benchmarks, so it looks too crowded, and anyway the user can see the benchmark name on the tooltip on hover (which is great!)
- [ ] I think right now for each benchmark, the bar for each solver are overlapped. This makes it hard to distinguish both, and to see the tooltip for both. Can we have them side by side, like how Matthias does?
image
image

- [ ] Instead of just having the License be a link, can we have the text say "License: MIT" etc? This can be hardcoded, we will quite rarely add a new solver to the page.

Thanks very much!

siddharth-krishna (commented):
Thanks, Jacek! This mostly looks good. But I think there might be a bug in the plotting code:

image
In this example, it says the GLPK time is 9s and SCIP is 600s, but if you look at the results CSV you'll see that all solvers TO on this benchmark. Perhaps the code is using the log runtime for GLPK? It should use log runtime only to plot the circles, but for the ratio calculation it should use non-log runtimes for both solvers.

Another thing I just realized: we should do something about TO values.. any ideas how we can represent them nicely? There's the case when the base solver (e.g. GLPK) TOs but the other one doesn't, and the case where the other solver TOs but the base solver doesn't, and then the case when both TO...

jacek-oet:
> In this example, it says the GLPK time is 9s and SCIP is 600s, but if you look at the results CSV you'll see that all solvers TO on this benchmark. Perhaps the code is using the log runtime for GLPK? It should use log runtime only to plot the circles, but for the ratio calculation it should use non-log runtimes for both solvers.

I fixed

image

> Another thing I just realized: we should do something about TO values.. any ideas how we can represent them nicely? There's the case when the base solver (e.g. GLPK) TOs but the other one doesn't, and the case where the other solver TOs but the base solver doesn't, and then the case when both TO...

@siddharth-krishna Should we add a maximum line to the chart so users can easily know the TO value? (red line in the chart)

siddharth-krishna (commented):
Thanks, Jacek! It looks much cleaner now 🤩 

A little issue is that now there are 2 tooltips when I hover, can we remove the little one? (If you can't reproduce it, I'm using Firefox on macOS)
image

Unfortunately, I like the new max time red line, but it's not enough because for the next benchmark run we will have 2 timeouts: 10h for the "real" size instances, and 1h for all others.. how about the following?

- [ ] We use `X` instead of `O` for the points where the base solver TOs. (Also add the same line from the home dashboard explaining this below the plot, please.)
- [ ] When the other solver TOs but the base solver doesn't, we do like you do now and make the bar go all the way to the max red line. But can we edit the tooltip so that instead of saying e.g. `Ratio: 1.1` it says `Ratio: N/A because scip TO`?
- [ ] When both solvers TO, we don't modify the bars and leave them at 1, so effectively they disappear as they are on the x-axis. This is like "no data", because we can't compute a ratio when both TO.

- [ ] Can we also add a header "Relative performance plot" above the graph, and the following sentence under the header to explain the plot?

> This plot shows the runtime ratios (relative speedup factors) for each benchmark instance with respect to the selected base solver's runtime. Ratios above 1 (bars above the x-axis) are instances where the base solver performs better, and ratios below 1 (bars below the x-axis) are those where the other solver performs better. Instances are sorted by the runtime of the base solver.

siddharth-krishna (approved):

PR #129, opened 2025-03-05
Website: improvements to filters, sorting, plots

jacek-oet:
- [x] Plots: zoom & pan features like plotly
- [x] 116 Results table: custom sorting for Solved Benchmarks and SGM Runtime
- [x] Filters: new filter design, easy ways to select all and select only one option
  - [x] Add an easy way to select all options in a category
  - [x] Add a way to reset all filters across all categories
  - [x] Reset filters when moving to a new dashboard
  - [x] Use URL query params to save filter state so that filtered results can be shared
- [x] Icons: use same kind of table sorting icons/arrows in all dashboards
- [x] Benchmarks: link to source / download each bench


siddharth-krishna (changes requested):
Thanks Jacek, this looks good overall, but a few questions/minor issues:

- [ ] The sorting icons disappear when I click on a column name on the Full Results page:
image
image

- [ ] "Add a way to reset all filters across all categories" -- you've checked this box as done, but I don't see this on the website? I was thinking about adding some kind of "Reset all filters" button/link -- not sure where it would look nice, please find a nice place to put it. :)
image

- [ ] I'm still seeing the bug where if I select only XS and then select the "all" checkbox on the main page, it doesn't go back to all results but is filtering to 74/137 instances. (Link to discord [message](https://discord.com/channels/1097854148965306470/1331265033090498590/1355181487414313030))

Thanks!

PR #131, opened 2025-03-10
Benchmarks: new GenX LP benchmarks without UC

danielelerede-oet:
Hi @siddharth-krishna , based on the GenX benchmarks on the jump-highs platform, I added here  a few LP GenX benchmarks removing Unit commitment.


siddharth-krishna (approved):
Thanks, Daniele! This looks good. If possible and not too much work, please do add a README and any configuration files etc you used to create the benchmarks so that they can be reproduced/updated when required. 🙏

danielelerede-oet:
Hi @siddharth-krishna @jacek-oet I added here 1 new GenX benchmark (the only one with non-linearized UC running with a credible time horizon - 30 days - after implementing Gabe's suggestion). In addition, I cleaned up the metadata commenting those UC problems with no solution, and I updated links and file names renaming the .lp files on the bucket  and the config files according to the usual nomenclature.

siddharth-krishna (approved):
It works now, thanks!

PR #142, opened 2025-04-04
Website: Sort in Benchmarks and SGM Runtime in Results table

jacek-oet:
…esults tabl


siddharth-krishna (approved):
Looks good, thanks Jacek!

A minor request since you're modifying this table in this PR, could you please use 2 decimal places for the values in SGM Memory and SGM Runtime columns?
image
So the above row would be:
```
1.70 (5145.00) | 12 % (9/74) | 1.36 (438.10)
```
Thanks!

siddharth-krishna (approved):

PR #145, opened 2025-04-08
Add contributors to metadata files and show on benchmark details page

danielelerede-oet:
Hi @siddharth-krishna I added the contributors to metadata files as discussed. I only have 2 doubts:
1) `Temoa Project` is listed as contributor for Temoa benchmarks, though they actually don't know about our project, so I'm not quite sure we should mention them as "contributors"
2) the `merge_metdata.py` script is not taking `Contributor` as an entry for the summary table, is it intentional?


siddharth-krishna (commented):
Thanks, Daniele!

1. Let's use the URLs to the GH repo for Temoa Project and Jump-dev platform
2. It's not intentional, it's a To Do. I can add that to this PR, and then perhaps @jacek-oet could help us by adding support on the website to display the Contributor(s) on the benchmark details page? Perhaps it can, for now, be a separate paragraph below the `Short description`, that looks like e.g. 
`Contributor(s)/Source: Gabe Mantegna, Princeton University; Daniele Lerede, Open Energy Transition`
image

danielelerede-oet:
@siddharth-krishna everything should be ready in agreement to what we discussed earlier!

siddharth-krishna (approved):
Brilliant, thanks Daniele! I'll update the metadata merging script now.

@jacek-oet could you please add task 2 above to your list? You can push directly to this PR branch. It would be good if URLs in the `Contributor(s)/Source` field appeared on the website as hyperlinks. Thanks!

siddharth-krishna (commented):
Thanks @jacek-oet , but I'm afraid the deployment doesn't work for me:
image

Can you reproduce the error? Otherwise I can try building locally.

PR #147, opened 2025-04-09
Implement reference benchmark

KristijanFaust-OET:
Introduces optional periodic reference benchmark runs.

If testing before merging this to `main`, you should add something like:
```
cd /solver-benchmark
git checkout reference-benchmark
cd ..
```
to the startup-script.sh after cloning the benchmark repo.

### DIFF
--- a/runner/run_benchmarks.py
+++ b/runner/run_benchmarks.py
@@ -2,10 +2,12 @@ import csv
 import gzip
 import json
 import os
+import re
 import shutil
 import statistics
 import subprocess
 import sys
+import time
 from pathlib import Path
 
 import requests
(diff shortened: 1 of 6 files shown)


runner/run_benchmarks.py
```diff
@@ -360,24 +446,65 @@ def main(
...
+                    print(
+                        f"Running reference benchmark with HiGHS binary (interval: {reference_interval}s)...",
+                        flush=True,
+                    )
+                    reference_metrics = benchmark_highs_binary()
+
+                    # Add required fields to reference metrics
+                    reference_metrics["size"] = "reference"
+                    reference_metrics["solver"] = "highs-binary"
+                    reference_metrics["solver_version"] = get_highs_binary_version()
```
siddharth-krishna, line 467: Could we call this once outside the loop and share the result for every iteration?

siddharth-krishna (approved):
Thanks Kristijan, this looks good! A minor request in comments. I don't have time to test it but I'm sure you did. :)

runner/run_benchmarks.py
KristijanFaust-OET: Addressed in: https://github.com/open-energy-transition/solver-benchmark/pull/147/commits/5d3b251d5d641529b96defac59b41ee30e8278c5

PR #150, opened 2025-04-14
Benchmark output capture

KristijanFaust-OET:


runner/run_solver.py
```diff
@@ -81,17 +81,25 @@ def get_duality_gap(solver_model, solver_name: str):
...
+
     solution_dir = Path(__file__).parent / "solutions"
     solution_dir.mkdir(parents=True, exist_ok=True)
-    solution_fn = solution_dir / f"{problem_file.stem}-{solver_name}.sol"
+
+    logs_dir = Path(__file__).parent / "logs"
+    logs_dir.mkdir(parents=True, exist_ok=True)
+
+    year_str = f"-{year}" if year else ""
+    output_filename = f"{Path(input_file).stem}-{solver_name}{year_str}"
```
siddharth-krishna, line 95: Minor request: can we use the solver version instead of the year? It's more future proof if we decide to go with some other strategy in the future like benchmarking every major version. Thank you!

siddharth-krishna (approved):

infrastructure/startup-script.sh
```diff
@@ -113,18 +117,101 @@ if [ "${ENABLE_GCS_UPLOAD}" == "true" ]; then
...
+    find /solver-benchmark/runner/solutions/ -type f -name "*.sol" | while read sol_file; do
+        filename=$(basename "${sol_file}")
+        compressed_file="${COMPRESSED_DIR}/solutions/${filename}.gz"
+
+        echo "Compressing ${sol_file} to ${compressed_file}..."
+        gzip -c "${sol_file}" > "${compressed_file}"
+
+        echo "Uploading ${compressed_file} to GCS bucket..."
+
+        # Check if file contains "gurobi" in the name
```
siddharth-krishna, line 175: Also, I just double checked: we can publish solution files, just not log files. Could you please make that change? 🙏

mattmilten: FYI: some solution files (for larger models, of course) can be huge, so you also need to ensure there is sufficient storage.

runner/run_solver.py
KristijanFaust-OET: Addressed in: https://github.com/open-energy-transition/solver-benchmark/pull/150/commits/1e8bc98d5c5e09528d9cf6ba673669dfc8c79eaf

PR #156, opened 2025-04-18
Benchmarks: add TIMES-GEO benchmarks

danielelerede-oet:
Adding TIMES-GEO benchmarks provided by Maurizio here :) should be real-size problems, they tested them with CPLEX and they were solved in 40-50 minutes


siddharth-krishna (approved):
Thanks, Daniele!

PR #157, opened 2025-04-18
Categorize benchmark sizes based on number of variables

KristijanFaust-OET:


siddharth-krishna (approved):
Thanks, Kristijan. I updated this PR with some changes:
- Updated your script to download benchmarks to the same path/filename as used by the runner to save re-downloading benchmark files
- I added a breakdown of integer vs continuous vars and updated the script to only calculate them for MILPs
- I made the script idempotent: don't download if already present, don't analyze if metadata already contains stats, etc. So that we can re-run this quickly if new benchmarks are added
- I renamed the metadata fields to use the more common abbreviation `Num.` instead of `N.`
- I updated the merge metadata script accordingly and ran it
- I marked all times benchmarks as realistic (they didn't have size categories previously), @danielelerede-oet when you're back please check that this is indeed correct.

Btw, this is the new breakdown of number of benchmark instances in each new size category:
```python
In [1]: import yaml

In [2]: from collections import Counter

In [3]: c = Counter()

In [4]: meta = yaml.safe_load(open('results/metadata.yaml'))

In [5]: for name, b in meta['benchmarks'].items():
   ...:     for s in b['Sizes']:
   ...:         c[s['Size']] += 1
   ...: 

In [6]: c
Out[6]: Counter({'M': 81, 'S': 18, 'L': 13})
```

Issue #159, opened 2025-04-18
[RFC] Should we use measured or reported solver runtimes?


siddharth-krishna:
The current benchmarking scripts [measure](https://github.com/open-energy-transition/solver-benchmark/blob/caab25dbb7da4973f7fedad2d62540b622078b25/runner/run_solver.py#L131-L135) the wall clock time (note: we are switching to `time.perf_counter` in an in-flight PR) taken by the call to `linopy.solvers.solve_problem`, which is a function in the Python library that acts as a uniform solver interface and handles parsing the solver responses and results. This means the measured runtime includes not just the time taken to solve the LP/MILP problem, but also the time taken by the solver to parse the input file and write the output file, or to check that the license (if any) is valid. Note that since all our benchmarks are LP/MPS files already, this measured time does not include the time taken by linopy to serialize the problem to a file.

An alternative is to use the **reported** runtime for solving from the solver. Some solvers expose this in their APIs:
```
Gurobipy: model.Runtime
PySCIPOpt: model.getTotalTime()
HiGHSpy: highs.getRunTime()
```
This runtime presumably excludes the time for input/output and license checks.

**Arguments for using measured runtime**:
1. It is closer to the end-user's experience: if a solver takes an extra 10s because it reads input files inefficiently or has to consult a remote license server, then the user will feel that the solver is 10s slower when they use it on their problems.
2. It does not require trusting that solvers do not (un)intentionally report inaccurate solving runtimes. (This can be mitigated to some extent by comparing reported runtimes to measured runtimes while analyzing benchmark results.)
3. If we use reported runtime, we will have to check that all solvers use a consistent definition of runtime (e.g. all of them exclude the time to parse inputs).

**Arguments for using reported runtime**:
1. It measures the core of what the solver does (solving the problem), and our performance history page reflects improvements to solver algorithms, not just faster input file parsers.
2. The measured runtime includes any time spent by `linopy` in the solver interface. For instance, if a solver does not have a Python API or linopy doesn't use one (e.g. CBC), then this runtime includes the time taken by linopy to parse the result/solution file.

We invite everyone to discuss the above in this issue. (@jajhall, @mattmilten, @pfetsch, @jjhforrest, and your teams, if you'll forgive the 2nd notification from me today, we would love to hear your thoughts on this as well.)

[...]

PR #161, opened 2025-04-23
Runner: use perf_counter and add reported runtimes

siddharth-krishna:
See the discussion in RFC issue #159

### DIFF
--- a/runner/run_solver.py
+++ b/runner/run_solver.py
@@ -1,7 +1,7 @@
 import json
 import sys
 from pathlib import Path
-from time import time
+from time import perf_counter
 
 import pandas as pd
 from linopy import solvers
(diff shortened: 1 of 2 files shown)


runner/run_solver.py
```diff
@@ -95,23 +96,41 @@ def get_milp_metrics(input_file, solver_result):
...
+            case "highs":
+                return solver_model.getRunTime()
+            case "scip":
+                return solver_model.getSolvingTime()
+            case "gurobi":
+                return solver_model.Runtime
+    except Exception:
+        print(f"ERROR obtaining reported runtime: {format_exc()}", file=sys.stderr)
+    return None
+
```
jacek-oet, line 132: Should we add a default case to raise an error for unsupported solvers?

siddharth-krishna: Great catch, thanks! I've fixed it

PR #164, opened 2025-04-28
Benchmarks: add resource adequacy problems

siddharth-krishna:
Thanks @lprieto1409 for contributing these benchmarks! A few small requests before we can merge this in:

- [ ] Can you make a smallest version that solves with Highs in 1hr?
- [ ] Can you add a readme with instructions of how to generate the benchmarks, which code and configs were used, what commit hash, what dependencies, what commands to run etc
- [ ] Could we rename the folder `benchmarks/resource_adequacy` to `benchmarks/<name of model>` if that makes sense?


results/metadata.yaml
```diff
@@ -1850,6 +1850,126 @@ benchmarks:
...
+    Kind of problem: Infrastructure
+    Sectors: Power
+    Time horizon: 179 days (~6 months)
+    MILP features: Unit commitment
+    Sizes:
+    - Name: 4296-997
+      Size: M
+      URL: https://storage.googleapis.com/solver-benchmarks/RA_DCOPF_2021_TE_10x_MILP_L2.lp
+      Spatial resolution: 997 nodes
+      Temporal resolution: 4296 (24 hours x 179 days)
```
siddharth-krishna, line 1967: Should this be `Temporal resolution: 1` because each time step is 1h long?

lprieto1409:
@danielelerede-oet I made the changes in temporal resolution, the folder has already been renamed to dcopf_carolinas

danielelerede-oet:
> @danielelerede-oet I made the changes in temporal resolution, the folder has already been renamed to dcopf_carolinas

Yeah sorry, I meant something different but I wasn't clear in asking that. I don't know if "dcopf_carolinas" is significant/in line with the other naming conventions, so what about renaming it to simply "resource_adequacy"?

lprieto1409 (approved):

jacek-oet:
@siddharth-krishna 
`DCOPF-Carolinas_1W` is missing `Model name`
```
DCOPF-Carolinas_1W:
    Short description: A tailored version of the DC optimal power flow (DC OPF) model
      configured for a one-week horizon with hourly resolution, using 2021 as the
      weather year for time series of load, solar, and wind generation. Resource adequacy
      is evaluated through slack variables, under transmission constraints, with support
      for nodal or zonal resolution.
    Model name:
    Version:
```

PR #165, opened 2025-04-28
Runner: allow specifying zone as tofu input for each VM

KristijanFaust-OET:


siddharth-krishna (approved):
Thanks, this worked! If you could also add your fix for the auto deletion to this PR we can merge it in

PR #166, opened 2025-05-01
Runner: handle OOM gracefully; update linopy version

KristijanFaust-OET:


siddharth-krishna (approved):
@KristijanFaust-OET thanks for this. I had to re-generate the fixed conda envs otherwise the `conda env create` command was erroring. I then successfully tested it on the benchmark that was OOMing, you can see logs on GCP under `20250510_102417_benchmark-instance-z2-m33`. If you can do a quick review of my changes, then we can merge this in. 🙏

PR #167, opened 2025-05-02
Preliminary results from the v1 run and fix some website issues

siddharth-krishna:
This PR:
- Brings in results for most existing benchmarks from the ongoing v1 run
- Fixes some bugs that this larger set of results revealed on the website:
  - Uses the Timeout value from the results in the home dashboard and SGM calculation
  - Fixes SGM calculation for non-OK statuses
  - Uses the new filters on all dashboards and removes the old Configuration top bar
  - Fixes Compare Solvers in the case when old solvers not run on L benchmarks
  - Fixes relative performance plot to the case where there are 2 timeouts
  - Adds a realistic breakdown to the benchmark summary plots
  - Fixes sorting of years in Performance History plot
  - Fixes Unknown Solver in tooltips and adds Status to tooltip


siddharth-krishna:
@jacek-oet these are the latest results from the GCP run. We are still awaiting some more results, but I thought I will make a PR for these so that we can test the website on them. I noticed a few bugs, could you please take a look at them and see if we can fix them? Thank you

- [ ] The years in the Performance History plots are not sorted
image
- [ ] When you filter to only "L" size, some data points show as "Unknown Solver" in the tooltip: 
image
- [ ] Can we also add `Status` to the tooltip above?
- [ ] Can we also add the size category to the tooltip? E.g. `Size: 3-1h (L)`
- [ ] Also when filtering to "L", and selecting the "Penalizing TO by a factor of" option, the SGM runtime value for gurobi goes up when you increase the factor from 5 to 6. However, gurobi solves all 15 benchmarks successfully and does not error on any of them, so the un-normalized SGM runtime in parenthesis should not increase!
image
image
 
- [ ] Compare solvers: if I select highs v1.10 and highs v1.9, I get `Application error: a client-side exception has occurred while loading solver-benchmark-4n9u52iv7-openenergytransition.vercel.app (see the browser console for more information).`
- [ ] Clicking a data point on the "Benchmarks" plot in the home dashboard does not take you to the benchmarks detail page for that benchmark
- [ ] Solvers page: relative performance plot looks a bit off. Can we rescale the y-axis so that all the lines fit in the graph? Why are some lines starting from the Xs and coming below?
image

jacek-oet:
Timeout has been added to the dashboard home page
image

siddharth-krishna:
Thanks, Jacek, I'm able to see the different results for the different tabs now! An additional minor request:

- [ ] Can you change the values in the "Configuration" bar to be c4-standard-2 for the short and c4-highmem-8 for the long tabs? And if you could please look up the number of vCPUs/cores/RAM for each of those GCP machine types and use them that would be gtreat.

> If the Runtime ratio (log scale) is less than 1, the lines will start from the X-axis and go below it
> ref: [mattmilten.github.io/mittelmann-plots/lpfeas-TULIP.html](https://mattmilten.github.io/mittelmann-plots/lpfeas-TULIP.html)

No, I mean a few of these lines that do not touch the x-axis:
image

jacek-oet:
> Compare solvers: if I select highs v1.10 and highs v1.9, I get Application error: a client-side exception has occurred while loading solver-benchmark-4n9u52iv7-openenergytransition.vercel.app (see the browser console for more information).

@siddharth-krishna The reason for the error is: Missing temoa-US_9R_TS (size: 9-12) for highs v1.9.

siddharth-krishna:
Thanks Jacek, I'll add the missing benchmark tonight.

siddharth-krishna:
> The reason for the error is: Missing temoa-US_9R_TS (size: 9-12) for highs v1.9.

@jacek-oet I took a look into this issue. It's actually not a missing result. The new benchmark run only runs the latest version of each solver on the L benchmarks to save time, because L benchmarks have a 10h timeout. So for the Compare Solvers page, what we need to do is to first filter the results to the benchmark-size instances that have been run on both solvers and then display the results/graph. Could you add that feature when you have some time, please? Thank you

siddharth-krishna:
Hi Jacek, I noticed another bug:

- [ ] Sorting by Solved Benchmarks is broken -- why is Gurobi treated as the lowest value?
image
image

Can you quickly check this when you have a minute today? Thank you 🙏

jacek-oet (approved):
LGTM

PR #169, opened 2025-05-04
Demonstrate a bug in SGM calculation

siddharth-krishna:


siddharth-krishna:
@jacek-oet I noticed another bug in the website, with the way that the SGM values are calculated. To demonstrate the bug, I made this branch where the results CSV file has only 3 benchmarks. I also created a notebook where I used our old python implementation of SGM as a reference to test the numbers against:
https://github.com/open-energy-transition/solver-benchmark/blob/89fe3a33f7b917a5ded8ff1f170d088d5a9077eb/tests/test-sgm.ipynb

If you look at the heading "Test on only 3 L benchmarks" in the above notebook, you'll see the values calculated by Python and can compare it to the ones calculated by the website. I think the problem might be the way rows with `Status != ok` are being handled, in particular `Status == ER`. Could you take a quick look at this please? This is first priority.

Also, I think this line:
https://github.com/open-energy-transition/solver-benchmark/blob/89fe3a33f7b917a5ded8ff1f170d088d5a9077eb/website-nextjs/src/components/admin/ResultsSections.tsx#L88
Should now be changed to
```javascript
            result.status !== "ok" ? result.timeout * xFactor : result.runtime,
```
so that it uses the Timeout column in the result CSV, because as in the example above, some non-OK statuses do not use the timeout value in the runtime column. So let's always use the Timeout column in any case when the Status is not `ok`.

siddharth-krishna:
Closing this PR as the bug was fixed in #167

PR #175, opened 2025-05-08
Website: Add links to logs & results for each benchmark instance to t…

jacek-oet:
…he benchmark details page

### DIFF
--- a/website-nextjs/src/components/admin/benchmark-detail/LogsAndSolutionsSection.tsx
+++ b/website-nextjs/src/components/admin/benchmark-detail/LogsAndSolutionsSection.tsx
@@ -1,0 +1,140 @@
+import { useEffect, useState, useMemo } from "react";
+import { useSelector } from "react-redux";
+
+import { IResultState } from "@/types/state";
+import { formatSolverWithVersion } from "@/utils/solvers";
+import { MetaDataEntry } from "@/types/meta-data";
+import Link from "next/link";
+
+const BASE_STORAGE_URL = "https://storage.googleapis.com/solver-benchmarks";
+
+const LogsAndSolutionsSection = ({
+  benchmarkDetail,
+  benchmarkName,
+}: {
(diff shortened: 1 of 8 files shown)


siddharth-krishna:
Thanks, Jacek, this looks good but when I was looking through the results last week I had another idea:

I think it might be useful to show the raw results of this particular benchmark in the benchmark details page. We can use the same table component as in https://solver-benchmark.vercel.app/dashboard/raw-result but filter the CSV results to this benchmark and have the following columns:
size, solver, solver version, status, termination condition, runtime, memory usage, objective value, max. int. violation (only if MILP), duality gap (only if MILP), log, solution
The last 2 columns can have download buttons similar to the buttons currently present in the "Instances" table.

This table can be present below the graph in the "Results on this benchmark" section. We can remove the MILP Features section as this information is present in the new table described above.

What do you think, @danielelerede-oet @lprieto1409 ?

danielelerede-oet:
Hi @siddharth-krishna @jacek-oet I think your solution works well. In general, I would prefer to have more information concentrated in a single place, so definitely good to remove the "MILP Features" table and to integrate them with the columns you mentioned in the same table.

siddharth-krishna:
Thanks, Daniele. In that case Jacek, would you mind implementing the table I had described [above](https://github.com/open-energy-transition/solver-benchmark/pull/175#issuecomment-2871164483) and removing the MILP Features table and the new "Logs and Solutions" section that you've built? Apologies for the double work.

siddharth-krishna:
Thanks, Jacek. Two minor requests please:

- [ ] Can we round runtime and memory to 2 decimal places?
- [ ] Can we add a first column "Instance" that gives the size instance name (e.g. 3-1h for `genx-6_three_zones_w_multistage`)

jacek-oet:
@siddharth-krishna 
I fixed.
image
image

siddharth-krishna (approved):
Looks good, thanks!

PR #176, opened 2025-05-08
Website: Fix filter issue

jacek-oet:
image


siddharth-krishna (commented):
Thanks Jacek, however, it still doesn't seem to work for me. To reproduce: go to Full Results page, click the Status column filter, uncheck All and try to check ER. It shows an empty table, and after that even if we select everything again the table stays empty:
image
image

Can you reproduce this? Thanks

jacek-oet:
@siddharth-krishna Added filter for Benchmark Set page


image

siddharth-krishna (approved):

PR #182, opened 2025-05-13
review short descriptions, add criteria for realistic

danielelerede-oet:
Hi @siddharth-krishna @lprieto1409 I added here the field to explain the motivation to consider a problem as Realistic when `Realistic: true`, reviewed the short descriptions and performed some small corrections in metadata.


siddharth-krishna (approved):
Thanks, Kristijan!

PR #186, opened 2025-05-15
Website: restrict performance history graphs to benchmarks solved by all versions

### LINKED ISSUE #183: Performance history misleading as L instances only run on latest solvers
Currently it looks like SGM runtime for most solvers drastically goes up for the latest version:

Image

But this isn't because the latest version of solvers is slower. It's because we have only run the L benchmarks on the latest versions, so the SGM value for those is much higher than that of previous versions. This can be seen by filtering out Ls:

Image

- [ ] Can we modify the code for this dashboard to first filter the results so that we compute SGM for each solver version on only the benchmark instances that have been run on all solver versions?
- [ ] Can we also add an SGM Mode dropdown so that the user can select different ways of computing SGM such as "Penalizing TO by factor"?

jacek-oet:
Fixes #183


website-nextjs/src/constants/index.ts
```diff
@@ -8,7 +8,7 @@ export enum ProblemSize {
 
 export const MaxRunTime = 36000;
 
-export const MaxMemoryUsage = 8192;
+export const MaxMemoryUsage = 7 * 1024;
```
siddharth-krishna, line 11: Should we remove this now? I thought we have another function somewhere that returns the max memory depending on the problem size S/M vs L?

siddharth-krishna (approved):
Thanks Jacek, this looks good! Curious why the performance history page doesn't have the SGM mode option of "on the intersection of solved benchmarks"? It's not a big deal though, I'm just curious if that's hard to implement for that page.

siddharth-krishna:
@jacek-oet thanks for the updates and comments. My responses:

> Should we update the memory usage of all benchmarks with a status other than 'ok' to the maximum memory usage (depending on the size S/M or L) on every dashboard page of the website? This would make all dashboard pages consistent.

Yes, I think that's a good idea. Similarly, for all non-`ok` statuses, the runtime should be the `Timeout` column value.

> Yes, I can. Will the 'Number of Benchmarks Solved' change based on the user's selected mode 'on the intersection of solved benchmarks'? If so, the number of solved benchmarks for all solvers in a given year would be the same.

Ah yes, I see. Okay let's leave it as it is, I don't think it's so important to have 'intersection of solved benchmarks' feature on this history dashboard.

siddharth-krishna (approved):
Thanks, Jacek!

PR #187, opened 2025-05-15
Update metadata with modelling framework and sectors

danielelerede-oet:
This PR makes the following changes:
- A new metadata field `Modelling framework` (e.g. TIMES) that is distinct from the `Model name` (e.g. eTIMES-EU).
- Renames `Technique` -> `Problem class`, `Kind of problem` -> `Application`
- `Sectors` metadata field is a comma separated list of sector names. `Power` sector is now called `Electric`.
- A new `Sector focus` distinguishes models that are `Power-only` vs `Sector-coupled`
- Updates `merge_metadata.py` to be more robust and to use the new fields above
- Updates the website to show the new metadata fields and to use a different filter logic for the `Sectors` field


siddharth-krishna:
Thanks, Daniele, this looks good. I'm wondering if we should also in this PR do some renaming of the metadata fields as suggested by Gurobi, what do you think?
```
Technique -> problem class
kind of problem -> application
```

- [ ] Reminder to self before merging this PR: we need to update `merge_metadata.py` to work with the renamed yaml fields
- [ ] We also need to update the website to show/use the new fields

danielelerede-oet:
> Thanks, Daniele, this looks good. I'm wondering if we should also in this PR do some renaming of the metadata fields as suggested by Gurobi, what do you think?
> 
> ```
> Technique -> problem class
> kind of problem -> application
> ```
> 
> * [ ]  Reminder to self before merging this PR: we need to update `merge_metadata.py` to work with the renamed yaml fields
> * [ ]  We also need to update the website to show/use the new fields

I was actually going to ask you about that, I'm doing it RN

benchmarks/genx-extended/metadata.yaml
```diff
@@ -1,12 +1,13 @@
...
     Short description: This is a one-year example which contains 15 zones. This example includes transmission expansion.
+    Modelling framework: GenX
     Model name: GenX
     Version:
     Contributor(s)/Source: Gabe Mantegna, Princeton University; Daniele Lerede, Open Energy Transition
-    Technique: LP
-    Kind of problem: Infrastructure
-    Sectors: Power
+    Problem class: LP
+    Problem class: Infrastructure
```
siddharth-krishna, line 9: Is it a mistake/typo that they're both the same key/field name? I think the second one should be `Application`?

siddharth-krishna (commented):
Thanks Daniele. So to confirm, we have `Sectoral focus` which is either `Power-only` or `Sector-coupled`, and we have `Sectors` which is a comma separated list of sectors. And the website filters for `Sectors` should display any problem where the selected sector appears (and possibly has more sectors). Right?

benchmarks/genx-extended/metadata.yaml
danielelerede-oet: Fixed!

siddharth-krishna:
To do on the website side:

- [ ] Update benchmark summary graphs
- [ ] Update Distribution of Model Features table: change the row headers to new fields, and use model framework instead of model name as column headers (if time, use model framework as first level column header, model name as second level column header).

jacek-oet (commented):
LGTM

jacek-oet (approved):
LGTM

siddharth-krishna (commented):
Thanks a lot Kristijan! The new filters look good to me, and the merge metadata script is much cleaner now ⚡ 

We still need to make a few small changes to the website:
- [ ] Update benchmark summary graphs to use the `Modelling framework` categories for the x-axis instead of `Model name`
image

- [ ] Update Distribution of Model Features [table](https://solver-benchmark.vercel.app/dashboard/benchmark-summary): use model framework instead of model name as column headers
- [ ] Also, in the same table, why is the Real size row full of zeros? Can we fix it?
image

- [ ] In the details page of a [benchmark](https://solver-benchmark.vercel.app/dashboard/benchmark-details/genx-elec_trex_co2), we need to add `Modelling framework` just after `Model name`

Let me know if you think you can do these easily, otherwise we'll ask Jacek for help. Thanks!

PR #195, opened 2025-05-21
Website: add problem size and realistic filters to Benchmark Set dashboard

KristijanFaust-OET:


siddharth-krishna (approved):
The code looks good to me, I'm merging it in to have the fix in production. @jacek-oet when you have time could you still review it and we can address issues in the next PR? 🙏

PR #196, opened 2025-05-21
Website: Relative performance plot (TOs and other errors)

### LINKED ISSUE #188: Relative performance plot: how do we handle TOs and other errors?
For example, check out [this URL](https://solver-benchmark.vercel.app/dashboard/solvers?kindOfProblem=Infrastructure%3BDC%2520optimal%2520power%2520flow%3BSteady-state%2520optimal%2520power%2520flow%3BOperational&modelName=TIMES-NZ&problemSize=L%3BM%3BS&sectors=Power%3BPower%252C%2520heating%2520biomass%252C%2520industry%252C%2520transport%3BUpstream%252C%2520Electric%252C%2520Transport%252C%2520Commercial%252C%2520Residential%252C%2520Industrial%3BPower%252C%2520heating%3BSupply%252C%2520power%252C%2520industry%252C%2520transport%252C%2520residential%252C%2520services%252C%2520agriculture%3BSupply%252C%2520power%252C%2520industry%252C%2520transport%252C%2520other&technique=LP%3BMILP&realistic=Realistic%3BOther) which is the Solvers page with `gurobi` selected and filtered to only Times-NZ model:
Image

The ratio between gurobi and glpk is NA, but why does the bar for glpk go down? If anything, it should go up because gurobi did better on this benchmark than glpk. The results for this times-nz-tui benchmark are:
```
times-nz-tui,2-24h,glpk,5.0,2020.0,warning,unknown,0.1073315780013217,295.94,,,,,3600.0,benchmark-instance-z-m34,20250429_090641_benchmark-instance-z-m34,2025-04-29 12:18:11.238400
times-nz-tui,2-24h,gurobi,12.0.0,2024.0,ok,optimal,308.7765187190089,652.508,1402087.105791976,,,307.4984369277954,3600.0,benchmark-instance-z-m34,20250429_090641_benchmark-instance-z-m34,2025-04-30 09:45:05.492636
```

We need to figure out what to do in the following cases:
- [ ] If the selected (base) solver has a non-`ok` status for a benchmark instance, then do we treat the base solver's runtime as the timeout value and compute the ratio? Or do we make the bar stay at zero? Or do we send the bar down to the bottom of the plot?
- [ ] If an other solver has a non-`ok` status, then do we treat the other solver's runtime as the timeout value and compute the ratio? Or do we send the bar up to the red dotted timeout line? I think the latter is what we do currently:
Image
- [ ] What do we do when both solvers have a non-`ok` status?

@mattmilten, @danielelerede-oet would love your thoughts here, thanks!

- [ ] Note for when we decide on a solution and implement it: the plot tooltip should also show the status of each solver. Also, at least for me I see 2 tooltips when I hover over a bar, which makes it hard to read, can we remove the smaller one?

Image

jacek-oet:
Fixes #188 

image
image
image

### DIFF
--- a/website-nextjs/src/components/admin/solvers/SolverSection.tsx
+++ b/website-nextjs/src/components/admin/solvers/SolverSection.tsx
@@ -36,11 +36,25 @@ const SolverSection = () => {
     return state.results.availableSolvers;
   });
 
-  const benchmarkLatestResults = useSelector(
+  const rawBenchmarkLatestResults = useSelector(
     (state: { results: IResultState }) => {
       return state.results.benchmarkLatestResults;
     },
   );
+
+  const benchmarkSuccessMap = new Map<string, number>();
+
+  // Count successful solves for each benchmark
+  rawBenchmarkLatestResults.forEach((result) => {
+    const key = `${result.benchmark}-${result.size}`;
+    benchmarkSuccessMap.set(key, (benchmarkSuccessMap.get(key) || 0) + 1);
+  });
+
+  // Filter results where all solvers succeeded
+  const benchmarkLatestResults = rawBenchmarkLatestResults.filter((result) => {
+    const key = `${result.benchmark}-${result.size}`;
+    return benchmarkSuccessMap.get(key) === availableSolvers.length;
+  });
 
   const [selectedSolver, setSelectedSolver] = useState("");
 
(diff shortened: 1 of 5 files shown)


siddharth-krishna (commented):
Thanks, Jacek, this looks good. A small request:

Can we add a legend somewhere, maybe the top right of the plot, that says:
> 🔻/🔺: base / other solver failed to solve in time limit
> ❌ : both solvers failed to solve in time limit

siddharth-krishna (approved):
Thanks!

PR #215, opened 2025-05-27
Website: Use log scale for runtime in Runtime vs Memory and Compare Solvers plots

### LINKED ISSUE #201: Use log scale for runtime in Runtime vs Memory and Compare Solvers plots
- [ ] Also, let's ensure that the axis starts at zero by default for all graphs plotting runtime, SGM runtime, SGM memory, number of benchmarks solved
- [ ] Compare solvers: Let's add the runtime and status of both solvers to the tooltip, e.g. `HiGHS v1.10.0: 4.3s (ok)`
- [ ] Compare solvers: Let's make the legend more explicit: instead of `ok-ok`, `ok-TO`, etc, let's use `gurobi ok - highs ok`, `gurobi ok - highs fail`, etc.

### LINKED ISSUE #211: Bug: sometimes filters are all unselected when navigating to a dashboard
This doesn't happen all the time, but sometimes going to a new dashboard by clicking the left side navbar link results in a dashboard where all filters are unselected and therefore 0 benchmarks are selected:

Image

Could we ensure that when a user navigates to a dashboard the default is for all filters to be selected? Thanks

jacek-oet:
Fixes #201 
Fixes #211

image
image
image


siddharth-krishna (approved):
Thanks for the updates! 2 tiny requests please:
- The status in this tooltip says ER for glpk but it should say TO. Can we use the status from the CSV file?
image

- Instead of the plot legend saying `glpk ok - highs TO` can it say `glpk ok - highs fail`?

siddharth-krishna (approved):
Thanks so much!

Issue #221, opened 2025-05-30
URL query params don't reflect filters on home dashboard


siddharth-krishna:
Selecting some filters used to change the URL, which is helpful to share a link to others so that they can view the same filtered results. Right now in production I don't see the URL query params changing, though the filtering still works:

![Image](https://github.com/user-attachments/assets/fb1cf19e-79a4-4f9e-b0f8-e79b1344b10f)

PR #224, opened 2025-06-02
Website: Consistent styles, filter fixes, remove passsword

### LINKED ISSUE #178: Website: consistent text and spacing styling everywhere using CSS classes
We need to have consistent fonts, font sizes, text colors, and spacing (margins / padding) across the website. One way to enforce this and to make it easy to update later is to use something like CSS classes for each of the main kinds of text (H1, H2, P, etc).

Here are some examples of where things can be improved:
- Landing page: Font of "Read more about our methodology" too small, "here" too big, make consistent with previous paragraph
Image
- Sometimes text is in grey, sometimes it is in dark blue. In some places the text looks smaller than others
Image
Image
- Both "Normalized SGM Runtime" and "Summary of Benchmark Set" are level 2 headings (H2), but have different sizes:
Image
Image
- The level 2 header  "Summary of Benchmark Set" has more padding below it compared to the level 1 header "Benchmark Set" (see pic above). Either they should have the same padding/spacing or the level 1 header should have more.
- Let's use the same capitalization style for all headers. "Summary of Benchmark Set" and not  "Summary of benchmark set".
- Let's have a consistent style for hyperlinks. In some places we're using underline, some places bold. It would also be good to have some hover effect, so that it is clear that it is a hyperlink.

@hebabadawy can you please ensure that the above are fixed when you update the Figma with the content from the https://solver-benchmark.vercel.app website? Once these are fixed on Figma, @jacek-oet can you please implement it on the website using style classes or something to ensure consistent styling everywhere? Thank you.

jacek-oet:
closes #178 #229 #221 

- More consistent text and spacing styling
- Fixes some typos
- Fixes filters jumping because "only" appears when hovering
- Fixes URL query parameters for filters
- Removes password protection


website-nextjs/src/components/admin/benchmark-detail/InstancesTableResult.tsx
```diff
@@ -244,9 +244,9 @@ const InstancesTableResult = ({
 
   return (
     <div className="py-2">
-      <div className="text-back text-2xl font-medium mb-2 mt-2 font-league pl-1.5">
+      <h5 className="text-navy font-medium mb-2 mt-2 font-league pl-1.5">
```
siddharth-krishna, line 247: Why h5 and not h2?

website-nextjs/src/components/admin/compare-solvers/SolverSelection.tsx
```diff
@@ -153,17 +149,14 @@ const SolverSelection = () => {
...
-        </div>
-        <p className="text-[#5D5D5D] text-sm sm:text-base 4xl:text-lg">
+        <h6 className="text-navy">Comparison</h6>
+        <p className="mb-6 mt-4 max-w-screen-lg">
           The benchmarks on the upper triangle of each graph are those where
           Solver 1 performs better, and those in the lower triangle are those
           where Solver 2 performs better. Click on any point in this graph to
           see details of that benchmark instance.
-          <p className="flex gap-1 items-center text-dark-grey text-sm">
+          <p className="flex gap-1 items-center text-navy text-sm">
```
siddharth-krishna, line 158: Why not define the text color in the style definition of `p`?

website-nextjs/src/components/admin/performance-history/NumberBenchmarksSolved.tsx
```diff
@@ -12,9 +12,9 @@ const NumberBenchmarksSolved = ({
 }: INumberBenchmarksSolved) => {
   return (
     <div className="pb-4">
-      <p className="text-navy font-bold leading-1.5 mb-3 4xl:text-xl">
+      <div className="tag-line font-bold mb-2">
```
siddharth-krishna, line 15: Why not use `<h2>` or `<h3>` here?

siddharth-krishna (commented):
Thanks, Jacek! This is a huge improvement. And thanks for fixing #229 !

I was curious why you were using h4/h5 etc. Can't we define h1/h2/h3 to be what we want? I don't think any page uses more than 3 levels of heading.

Can we also use a uniform text width? It looks like in  these two dashboards, the paragraph under "Relative performance plot" has a smaller fixed width and the paragraph under "Comparison" uses all available width.
image
image
Performance History text has one paragraph with fixed width and the other without?
image

Links: can we have a consistent link style? The landing page link underlines on hover, but the dashboards links are always underlined:
image
image

website-nextjs/src/pages/dashboard/home.tsx
```diff
@@ -66,7 +66,7 @@ const LandingPage = () => {
                 use reported or measured runtime{" "}
                 <a
                   href="https://github.com/open-energy-transition/solver-benchmark/issues/159"
-                  className="underline"
+                  className="hover:underline underline-offset-4"
```
siddharth-krishna, line 69: Instead of adding these classes to every link on every file, can't we add these to the definition of `a` in `website-nextjs/src/styles/globals.css`?

siddharth-krishna (commented):
Thanks, Jacek, this is awesome. Just a few minor things:

I'm seeing a horizontal scroll bar in the filters, but only in the main dashboard. Can you see it too? If so, is it easy to fix?
image

Can we have some standard spacing between two `<p>` paragraphs, like in this dashboard, or in the "Ranking" section of the main dashboard?
image

Issue #202, opened 2025-05-22
History dashboard improvements


siddharth-krishna:
- [x] No. benchmarks solved plot: set y-axis to be always 0 to total number of benchmarks, use dotted line to mark max possible value
- [ ] SGM runtime/memory: let's modify the code so that instead of normalizing all values to the best solver version across all solvers and all years, we normalize the values in each year by the best solver version in each year. So, for example, the SGM runtime plot should look like the following: (also let's add the dotted line at y=1 and labels on each data point that show the normalized values in the form of `4.2x`)
    ![Image](https://github.com/user-attachments/assets/325a911b-e670-4518-8ad8-7bf8d4893b60)
- [ ] Let's add a row of plots that show performance evolution for each solver individually. This plot, see an example below, shows in blue bar chart the number of unsolved problems in the benchmark set (after filters are applied), and in a red line chart the SGM runtime of each year's solver version normalized to the first year's version that we have data for.
    ![Image](https://github.com/user-attachments/assets/9cc6cccc-b602-4ed3-b1bc-6c5f9cc0991b)
    We don't need to use the colors in the above picture, let's use the same style and colors of the other existing plots on the History dashboard. Also, the line graph doesn't need to look like it does in the above picture from Gurobi, it can look like the ones we have currently.

### Old idea (discarded)

One idea for the SGM runtime/memory graphs: have a row of bar charts, one per solver, sharing the y-axes.

![Image](https://github.com/user-attachments/assets/af375db0-a4aa-448c-995f-329e58d80add)

This plot was generated with the python code:
```python
import pandas as pd
import matplotlib.pyplot as plt

# Create a DataFrame from the CSV data
data = {
    'solver': ['glpk', 'highs'],
    '2020': [13.4, 1.43],
    '2021': [15.6, 1.54],
    '2022': [12.1, 1.32],
    '2023': [10.3, 1.0]
}
df = pd.DataFrame(data)
# Set the years as the index for easier plotting
years = df.columns[1:]
df.set_index('solver', inplace=True)

# Create a figure and axes for the bar charts
fig, axes = plt.subplots(1, len(df.index), figsize=(12, 6), sharey=True)

# Calculate deltas for all solvers
deltas_dict = {}
for solver in df.index:
    deltas_dict[solver] = df.loc[solver].diff().fillna(0)
    deltas_dict[solver] = deltas_dict[solver] * 100 # Make it percentage

# Determine limits for the secondary y-axis based on deltas
[...]

PR #227, opened 2025-06-03
Website: add max value to number of benchmarks solved plot

KristijanFaust-OET:
Issue: https://github.com/open-energy-transition/solver-benchmark/issues/202


jacek-oet (approved):
LGTM

PR #233, opened 2025-06-05
Benchmarks: update and fix metadata (modelling framework, application, etc.)

danielelerede-oet:
Fixes the mismatch in Modelling framework for Sienna benchmarks, fills in missing realistic motivations, updates application to say capacity expansion explicitly, etc.


siddharth-krishna (approved):
Thanks, Daniele. I reverted the name changes as discussed and will open a separate PR soon.

PR #240, opened 2025-06-09
Benchmarks: fixes to metadata application field

danielelerede-oet:
@siddharth-krishna here I'm fixing some metadata fields which should not be disruptive for the website, just helpful to make everything more uniform and understandable


siddharth-krishna (approved):
I was wondering about this! Thanks for the fix :)

PR #244, opened 2025-06-12
Website: Add relative runtime bar plots to website

### LINKED ISSUE #243: Add relative runtime bar plots to website
As we are drafting a new Key Insights page that will have more accessible results and analysis, one plot that we are planning to use in multiple places is a new relative runtime plot, similar to what we have on the [solvers](https://solver-benchmark.vercel.app/dashboard/solvers) page, but a bit simpler:

Image

The python code that produced the plot above can be found in the "Relative runtime bar plots" section of [this](https://github.com/open-energy-transition/solver-benchmark/blob/key-insights-poc/pocs/key-insights.ipynb) notebook. I couldn't implement all the features in python, so please add the following features when implementing it in JS:

- [ ] Hovering over a bar in the plot brings up a tooltip that has the following information, e.g.
    ````
    Solver: highs v1.10.0
    Runtime: 25 min
    Memory: 350 MB
    ````
- [ ] The x-axis tick labels should be customizable. In the example above we show only the name of the benchmark and the runtime of the fastest solver, but in some other places we want to additionally show some other metadata like num vars/constraints. Alternatively, if easier, we can show the custom information in a tooltip when hovering over the tick label.
- [ ] Let's reduce the size of the bars that correspond to a non-ok status (in the example above these are the ones with the red text on top of the bars) so that they are also plotted to the y=1 dotted line. This is so that they do not attract too much attention, or give people the wrong impression that the solver took an infinite amount of time on those cases. But please note that those should still be sorted to be on the left of any bars that have ok statuses, so you need to modify the values after sorting but before plotting, as in my python notebook.

Can we add a function/module to the website that produces such plots, given some results data for a given set of benchmarks? I expect the input data to have 1 or more benchmarks, and have at least the columns `Benchmark, Solver, Runtime (s), Status, Timeout`. (See my python code in my [notebook](https://github.com/open-energy-transition/solver-benchmark/blob/key-insights-poc/pocs/key-insights.ipynb) for example.) Once we have code to do this in a re-usable way, we can use the same function/module to add this plot to multiple places in the website:

- [ ] Let's replace the Runtime vs Memory plot in the benchmark details page for a benchmark (e.g. https://solver-benchmark.vercel.app/dashboard/benchmark-set/genx-elec_trex_co2) with this new plot, containing all the size instances for that benchmark. The x-axis tick label can be the size instance name and the fastest solver runtime, and hovering over a bar can show all the information shown currently in the Runtime vs Memory plot's tooltip.
- [ ] Let's also add this plot to the current Main Results dashboard. But instead of plotting individual benchmarks, let's use as input the unnormalized SGM runtime values for each solver. This should result in a plot that looks like this:
    Image
    The plot should be dynamic and use the same data as used in the table, so changing filters should change the data in the plot. The x-axis tick label should say what the fastest solver is and the average runtime in human-readable format. The tooltip when hovering over each bar should contain e.g.
    ````
    Solver: highs v1.10.0
    Average runtime: 7 hrs
    Benchmarks solved: 5/15 (33%)
    ````

jacek-oet:
Fixes #243, #205 
image
image
image
image

### DIFF
--- a/website-nextjs/src/components/admin/ResultsSections.tsx
+++ b/website-nextjs/src/components/admin/ResultsSections.tsx
@@ -360,8 +360,10 @@ const ResultsSection = ({ timeout }: ResultsSectionProps) => {
             ) + 1,
           solver: solverData,
           unnormalizedData: {
-            runtime: runtimeSgm,
-            memoryUsage: memorySgm,
+            runtime: calculateSgm(getRelevantResults(solverData, "runtime")),
+            memoryUsage: calculateSgm(
+              getRelevantResults(solverData, "memoryUsage"),
+            ),
           },
           version: getHighestVersion(solverVersions[solverData]),
           memory: `${memorySgm === minMemory ? "<b>" : ""}${roundNumber(
@@ -606,6 +608,7 @@ const ResultsSection = ({ timeout }: ResultsSectionProps) => {
         </div>
       )}
       <SgmRuntimeComparison
+        timeout={timeout}
         sgmData={tableData}
         uniqueBenchmarkCount={uniqueBenchmarkCount}
         uniqueLatestBenchmarkCount={uniqueLatestBenchmarkCount}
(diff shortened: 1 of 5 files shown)


siddharth-krishna (commented):
Thanks, Jacek, this is great! Just a few minor changes please:

- [ ] The plot in the Main dashboard should use the unnormalized SGM runtime when calculating average runtime in the tooltip and the x-axis label, but right now it is using the normalized SGM runtime. (e.g. it is right now 17s for glpk in your screenshot but it should be 26 min)

PR #246, opened 2025-06-13
Website: Replace Model filter with Modelling framework filter

jacek-oet:
image
image


siddharth-krishna (approved):
Thanks, Jacek! A hopefully-quick related request: can we re-order the filters in all dashboards to be in this order: modelling framework, application, problem class, problem size, realistic, sectoral focus, sectors

Thanks!

siddharth-krishna (approved):

PR #252, opened 2025-06-17
fix: timeout-bars-in-runtime-plot

jacek-oet:


siddharth-krishna (approved):
Thanks, Jacek, this looks good. In your example though, why are all the bars so small? Can't we scale the y-axis so that it goes from 0 to 20 or something since no bar goes above 13.6? Thanks.

jacek-oet:
@siddharth-krishna I updated.

siddharth-krishna (approved):
Thanks!!

PR #263, opened 2025-06-20
Website: Key Insights: solver evolution

jacek-oet:
image

### DIFF
--- a/website-nextjs/src/components/admin/performance-history/SolverEvolutionSection.tsx
+++ b/website-nextjs/src/components/admin/performance-history/SolverEvolutionSection.tsx
@@ -17,12 +17,16 @@ interface ISolverEvolutionSection {
   solverYearlyMetrics: ISolverYearlyMetrics[];
   numSolvedBenchMark: ISolverYearlyChartData[];
   totalBenchmarks: number;
+  title?: string;
+  description?: string;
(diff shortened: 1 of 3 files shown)


jacek-oet:
> One request: for the "performance evolution" plots (both on this page and the history dashboard), can we use the same y-axis range for speedup? Let's check the min and max values across all solvers and use a slightly bigger range. E.g. 0.9 to 1.7 for all solvers. Otherwise right now it looks like a huge drop in performance for HiGHS but it's actually just going from 1 to 0.97 which isn't that big of a slowdown, compared to other solvers.

image
I updated 


> Another thing: is there a bug in the tooltips? It says speedup of 1.00x for Highs for the first 3 versions, but the graph looks like it's going up and then down:

It shows that because the current values are:

1
1.0034459461967835
0.9984387869082831
0.9746075128934903

These values are currently rounded to two decimal places, That’s why you see the graph looks like it goes up and then down.


> Sorry, another thing, can we add the max number of benchmarks red dotted line to this plot too? And make sure the y-axis range on the left goes from 0 to a bit more than the max number? Thanks

I updated 
image
image

siddharth-krishna (approved):
Thanks Jacek, that all looks great! 🙏

PR #266, opened 2025-06-21
Website: add a metadata template

danielelerede-oet:
@siddharth-krishna I developed this template for metadata, so I can use it for the presentation at EIS


siddharth-krishna (approved):
Looks good, thanks! Should we also add a section to the top of the main README file about our benchmarks set, and how to contribute new ones?

siddharth-krishna:
The merge_metadata.py script works on all files called `benchmarks/metadata*.yaml`... maybe we should rename this file as `benchmarks/template.yaml`?

danielelerede-oet:
> Looks good, thanks! Should we also add a section to the top of the main README file about our benchmarks set, and how to contribute new ones?

I opened an issue for that, but we can also add it to the README, sure

danielelerede-oet:
> The merge_metadata.py script works on all files called `benchmarks/metadata*.yaml`... maybe we should rename this file as `benchmarks/template.yaml`?

That's true, what about `_template_metadata`?

siddharth-krishna:
Thanks, and sure that works!

PR #289, opened 2025-06-27
Website: History fix y axis range for speed up

### LINKED ISSUE #287: Add units to runtime & memory in benchmark details page table

### LINKED ISSUE #288: History: fix y-axis range for speed-up to be the same for all solvers
We need the y-axis on the right side to have the same range for all solvers:

Image

Image

jacek-oet:
closes #287
closes #288 

image
image
image


siddharth-krishna (commented):
Thanks, Jacek, this looks good. A quick request, can the "Max: 105" text above the red line be on the left instead of the right? Because it is related to the left y-axis, not the right one. Thanks 🙏

PR #304, opened 2025-09-10
Website: Update component styles to use navy background and white text

jacek-oet:
- [x]  In the key insights page, when scrolling from the bottom to the top in Firefox, the correct subtitle is not selected in the left bar "On this page"
- [x]  The performance evolution plot in the What is feasible for open source solvers? section has a red line for the Max, can it be the same green color as the bars?
- [x]  The tooltips in the filters have a nice dark background, but can we also make the tooltips in the main section have same background color? E.g. the one after "SGM (?)" in the main results page
- [x]  Can we change the FAQ item component to accept any HTML component as argument so that the answer can have proper links, bold/italic text, etc?
- [x] Charts

image
image
image

Removed max-height for filter section in mobile device.
image

And collapse filters by default in mobile resonsive.
image

In mobile, use diagonal x-axis labels instead.
image

Reduce the footer padding
image

Updated Tooltip background
image
image
image
image
image

image
image
image
image
image
image
image
image
image


siddharth-krishna (approved):
Looks good, thanks! Is the deployment failing because I've added you to Vercel as a team member with Viewer permissions? 🤔  that's annoying!

PR #306, opened 2025-09-15
Add benchmark models

nour-boulos:
This PR adds 3 benchmark models of ZEN-garden.
1. zen-garden-eur-PI-annual-emission-limit: the same as PI_OEB_100 but with an annual emission target that results in the same cumulative emissions as the carbon emission budgets. This should simplify the problem as inter-annual time coupling is not required.
2. zen-garden-eur-PI-constrained-expansion: the same as PI_OEB_100 but with constrained technology expansion (adds a lot of complexity through inter-annual time coupling).
3. zen-garden-eur-PI-no-storage: the same as PI_OEB_100 but without storage technologies.

### DIFF
--- a/benchmarks/zen_garden/metadata.yaml
+++ b/benchmarks/zen_garden/metadata.yaml
@@ -1,7 +1,7 @@
 ## https://github.com/ZEN-universe/ZEN-garden
 benchmarks:
   zen-garden-eur-PI:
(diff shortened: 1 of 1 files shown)


danielelerede-oet:
Thank you for the new benchmarks @nour-boulos . For the sake of simplicity, I moved the changes to the metadata to the older PR so that it's easy to manage and merge, as some merge issues seem to be arisen here. Therefore, I will close this PR.

PR #313, opened 2025-09-29
Website: Add link to benchmark in TableResult and enable showAllRows

jacek-oet:


siddharth-krishna (commented):
Thanks, Jacek, but I'd like some changes please:

Image

- Some columns are too small in width, I can't see many column headers (Instance shows as `I...` and Status is not visible at all)
- This table is still paginated and says "Showing 1 to 10 of 69". Can we have it show all rows with a scroll bar?

PR #314, opened 2025-10-03
Benchmarks: add IESA-Opt-NL model instances

ahfattahi:
Adding IESA-Opt for benchmark
 

- [x]  The benchmark name and size instance name follow the conventions indicated in the template
- [x]  I have tested that this model instances can be solved to optimality in 1 minute to 30 hours with Gurobi 12.0 solver, using options below, on a Lenovo P620 work station machine.

Options: 
- option "GUROBI 12.0".Scale := 1 ;
- option "GUROBI 12.0".Presolve := 1 ;
- option "GUROBI 12.0".Presolve_aggregation := 2 ;
- option "GUROBI 12.0".Presolve_aggregation_fill := 5 ;
- option "GUROBI 12.0".Presolve_row_reduction := 1 ;
	
Lenovo P620 spec:
- Processor	AMD Ryzen Threadripper PRO 3995WX 64-Cores (2.70 GHz)
- Installed RAM	512 GB (512 GB usable)

### DIFF
--- a/benchmarks/iesa-opt/metadata.yaml
+++ b/benchmarks/iesa-opt/metadata.yaml
@@ -17,7 +17,7 @@ benchmarks:
     Sizes:
     - Name: 1-3h
       Size: M
-      URL: https://storage.googleapis.com/solver-benchmarks/instances/IESA-Opt-NL-1n-3h-1y.mps.gz
+      URL: https://storage.googleapis.com/solver-benchmarks/instances/IESA-Opt-NL-1-3h.mps.gz
       Temporal resolution: 2920 time slices
       Spatial resolution: 1 node
       Realistic: true
@@ -26,7 +26,7 @@ benchmarks:
       Num. variables: 493813
     - Name: 1-1h
       Size: L
-      URL: https://storage.googleapis.com/solver-benchmarks/instances/IESA-Opt-NL-1n-1h-1y.mps.gz
+      URL: https://storage.googleapis.com/solver-benchmarks/instances/IESA-Opt-NL-1-1h.mps.gz
       Temporal resolution: 8760 time slices
       Spatial resolution: 1 node
       Realistic: true
@@ -35,7 +35,7 @@ benchmarks:
       Num. variables: 1107013
     - Name: 10-1h
       Size: L
-      URL: https://storage.googleapis.com/solver-benchmarks/instances/IESA-Opt-NL-10n-1h-1y.mps.gz
+      URL: https://storage.googleapis.com/solver-benchmarks/instances/IESA-Opt-NL-10-1h.mps.gz
       Temporal resolution: 8760 time slices
       Spatial resolution: 10 nodes
       Realistic: true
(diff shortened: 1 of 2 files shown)


danielelerede-oet:
Hello @ahfattahi and thanks for contributing! I would only need you to provide me another link for the benchmarks as the one in the metadata is expired (I will move the files to our repository once I download them, so that you don't have to necessarily store them on a permanent location). Also, could you please provide a bit of up-to-date documentation on the model? I've seen the main repository of IESA-Opt is not maintained since a few years.
Also, would you have the possibility to easily generate different size instances of the same model (i.e. with different time and spatial resolution)?
Thank you again!

siddharth-krishna:
Hi @ahfattahi thank you very much once again for your contribution to the Open Energy Benchmark. We apologize for the delay in getting back to you, but unfortunately your file share link has expired. If you can find the time to send us a new link, we are now preparing for the v2 benchmark run and can include your benchmarks in it. Thank you!

ahfattahi:
Hi @siddharth-krishna and @danielelerede-oet, apologies for my rather late reply. I was in preparation of variety of .mps files with different temporal and regional resolution. Unfortunately, we do not have an updated github repository explaining the model. However, there are several papers using this model. I provided the link to some of them in the yaml file. 
Please note that the download link is valis only until 8th of January 2026. The link includes all the mps files. 
Thanks in advance for all your efforts.

ahfattahi:
Hi both, thank you for your swift responses and improvements to the PR. 

@siddharth-krishna I have updated the PR's description. Please let me know if further information from my side is needed. I also added a description on how to get the MPS files. However, the database of different geographical resolutions are not available on Github yet as we are writing a paper on it. It will be published once the paper is out.
On model runtimes, it is a pity that you are not running the large-scale models. Particularly, I think it is useful to look at the performance of new algorithms such as PDHG, on these large-scale LPs. Anyways, I suggest you run the 5n-3h-1y, 20n-3h-1y instances as they have lower temporal resolution and might end up solved below your current time-limit. 

@danielelerede-oet, I was not sure what "realistic" would mean exactly. My interpretation was that a case is "real" if it is a representation of a real case, and not a dummy model of an imaginary energy system. One would argue that all energy models are not real as they are a simplification of reality. For many years Dutch energy policy (and many others) was driven by single node energy models. 
Here, you know best what "real" would mean and I trust your interpretation. Please correct the yaml file as you see fit.

siddharth-krishna:
@ahfattahi thank you for the updates, and apologies for the slow reply as we work on getting release v2 out!

Indeed, we hope to benchmark algorithms like PDHG in a future release, so I suggest we keep the larger instances in the PR as well -- we now have a way to "skip" executing benchmarks in this release -- and keep them skipped until we update the set of solvers / time limit.

Regarding barrier / crossover: our current policy is to run all solvers with their default options, so we measure the total time taken by the solver, and it is up to the solver to decide whether to run crossover or not. We are considering for a future release to allow solver developers to provide different "variants", each essentially a set of solver options that are chosen to perform well on a subset of energy models.

PR #318, opened 2025-10-15
Website: Fix virtual row positioning in tan stack table component

jacek-oet:
image
image


siddharth-krishna (approved):
Thanks, it looks good now!

PR #319, opened 2025-10-24
Website: Add GitHub link and adjust layout

jacek-oet:


siddharth-krishna (approved):
Thanks for the quick turnaround, Jacek!

PR #326, opened 2025-11-06
Infrastructure/Runner: benchmark campaign setup

madhukar93:


siddharth-krishna (commented):
Thanks a lot! This is a huge amount of work and should really help us with the v2 run in Dec. 🙏 

My review here since it's hard to comment on notebooks in the diff view:

`notebooks/allocate-benchmarks-to-vms.ipynb`:
- [x] Why is the "Generate Optimal Allocation" section empty?
- [x] Why are we using a 1h `MAX_RUNTIME_PER_VM_SECONDS`? It leads to a lot of errors in the output of the bin packing
- [x] Is there a way to set the max number of VMs to 40? That's the number of Gurobi licenses we have, so we can't run more than 40 VMs at the same time.

`notebooks/run-and-observe-benchmarks.ipynb`:
- [x] These cell numbers aren't visible in the notebook, perhaps use numbered sections instead?
Image

- [ ] Why are we calling the vm script with `!python $VM_SCRIPT` instead of `%run` or better still importing the functions from the script and calling them in python cells?
- [ ] What's the GCS results?

But great work overall, looking forward to using the new notebooks and utils in my next run 🚀

madhukar93:
thanks for the detailed review

> Is there a way to set the max number of VMs to 40? That's the number of Gurobi licenses we have, so we can't run more than 40 VMs at the same time.

Makes sense to have that config.

> Why are we using a 1h MAX_RUNTIME_PER_VM_SECONDS? It leads to a lot of errors in the output of the bin packing

An artifact of testing that got left in. Yes, 1h causes overflows and doesn't make sense for practical cases.

> Why are we calling the vm script with `!python $VM_SCRIPT` instead of `%run` or better still importing the functions from the script and calling them in python cells?

We can do imports from the gcloud wrapper script instead of shell execution. The notebook is an odd mix of shell and python currently because it is a combination of disparate scripts that we used to run the HiGHS hipo benchmarks, so forgive me if the code doesn't read very well right now.

> What's the GCS results?

Can you clarify?

notebooks/run-and-observe-benchmarks.ipynb
```diff
@@ -37,15 +36,23 @@
     "run_id = \"20251106_153156_batch\"\n",
     "```\n",
     "\n",
-    "2. Use **Cell 2** to query VM status while running\n",
-    "3. Use **Cell 3** to download results from running VMs\n",
-    "4. Use **Cell 4** to download from GCS\n",
-    "5. Use **Cell 5+** to combine and analyze results"
+    "2. Use *section 1.Configure variables for the Benchmarking Campaign* to point the notebook at your session\n",
+    "3. Use *section 2.Observe the Running Benchmark Session* to download result CSVs and logs to observe the progress of the Benchmarking Session"
```
siddharth-krishna:
```suggestion
    "2. Use section *1. Configure variables for the Benchmarking Campaign* to point the notebook at your session\n",
    "3. Use section *2. Observe the Running Benchmark Session* to download result CSVs and logs to observe the progress of the Benchmarking Session"
```

notebooks/allocate-benchmarks-to-vms.ipynb
```diff
@@ -11,22 +11,36 @@
...
-    "- **No Multiplier**: Uses exact runtime values from the results CSV (not scaled)\n",
     "- **Algorithm**: Balanced partition algorithm for optimal load distribution\n",
-    "- **Machine Separation**: L-size benchmarks on highmem VMs, S/M-size on standard VMs"
+    "- **Machine Separation**: L-size benchmarks on highmem VMs, S/M-size on standard VMs\n",
+    "\n",
+    "## Workflow\n",
+    "1. Configure how long you want the benchmarks to run using `MAX_RUNTIME_PER_VM_SECONDS`, smaller values will result in more VMs being used. You can also set `MAX_NUM_VMS` to limit the number of VMs used. Allocation will error if the configuration is unfeasible.\n",
+    "2. Filter out the benchmark results you want to re-run from the previous run in `results/benchmark_results.csv`.\n",
+    "3. Run the playbook to generate allocation config in infrastructure/benchmarks/<run_id> directory.\n",
+    "4. Launch the benchmarks using (Opentofu)[https://opentofu.org/]. Read `infrastructure/README.md` for details of the infrastructure orchestration setup.\n",
```
siddharth-krishna:
```suggestion
    "4. Launch the benchmarks using [OpenTofu](https://opentofu.org/). Read `infrastructure/README.md` for details of the infrastructure orchestration setup.\n",
```

Issue #325, opened 2025-11-06
Website: solvers page should mention solver version


siddharth-krishna:
Can we please include the version of the solver in the dropdown in the solvers page?

<img width="1464" height="489" alt="Image" src="https://github.com/user-attachments/assets/bb3491ac-1328-4865-8c90-69e478dfeafa" />

Right now it's not clear which version's data is being shown in the "Relative performance plot" below. Can we have each item in the dropdown look like "glpk v5.0"?

Issue #324, opened 2025-11-06
Website: fix breadcrumb


siddharth-krishna:
The breadcrumb for https://openenergybenchmark.org/dashboard/benchmark-set is "Benchmark Set":
<img width="720" height="231" alt="Image" src="https://github.com/user-attachments/assets/4546347a-7727-4529-af5b-82676e227d12" />
But the breadcrumb for a particular benchmark, e.g. https://openenergybenchmark.org/dashboard/benchmark-set/genx-elec_trex_uc, is "Benchmark Details > genx-elec_trex_uc". Can we use Benchmark Set here too?
<img width="807" height="229" alt="Image" src="https://github.com/user-attachments/assets/ebb66910-ea24-4402-b3e2-915526a10e07" />

Issue #323, opened 2025-11-06
Website: use URL query params for all dashboards with selectors


siddharth-krishna:
Can we please use URL query params for dashboards such as the Compare Solvers page where the output depends on the solvers selected?

<img width="1491" height="469" alt="Image" src="https://github.com/user-attachments/assets/c9f1674d-61b4-4c73-889e-9470f1b60eec" />

This will let us share a particular comparison between 2 solver versions via URL.

Please also do this for:
- Main Results: Short vs Long
- Solvers: which solver is selected
- Performance History: Individual Solver Performance Evolution (can we also have a bookmark so that it scrolls down to this section if URL is shared for performance evaluation of e.g. SCIP?)

PR #328, opened 2025-11-12
Website: Bulk Benchmark download & UI issues

jacek-oet:
- Implemented bulk benchmark download

Also closes:
- #323
- #324
- #325 

image
image
image


siddharth-krishna (commented):
Thanks, Jacek, this is cool! Could you please also add URL query params to these pages:

> Solvers: which solver is selected
> Performance History: Individual Solver Performance Evolution (can we also have a bookmark so that it scrolls down to this section if URL is shared for performance evaluation of e.g. SCIP?)

siddharth-krishna (approved):
Thanks Jacek 🚀

PR #329, opened 2025-11-14
Add ethos.fine multi regional example to solver benchmark

JohannesBehrens:

PR #333, opened 2025-11-18
Benchmarks: add ethos.fine multi regional example

JohannesBehrens:
This is the new PR, as mentioned in #329. I added the requested description of the regional focus of the example.
I also reviewed the other examples. However, most of them are very simple scripts designed to demonstrate how the fine code works. I don't think they're complex enough to serve as a benchmark.

### DIFF
--- a/benchmarks/times/metadata.yaml
+++ b/benchmarks/times/metadata.yaml
@@ -315,7 +315,7 @@ benchmarks:
     Application: Infrastructure & Capacity expansion
     Sectoral focus: Sector-coupled
     Sectors: Supply, Electric, Transport, Industrial, Other
-    Time horizon: Single stage
+    Time horizon: Single-stage
     MILP features: None
     Sizes:
     - Name: 2-24h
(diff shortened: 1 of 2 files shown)


danielelerede-oet:
Thanks @JohannesBehrens ! I have reported here the same changes made on the older PR to align with website-friendly metadata fields and update the URLs after storing the files on our repository. 

I'm also reporting here the same request as in the other PR: as I have also noticed a few FINE models [examples](https://github.com/FZJ-IEK3-VSA/FINE/tree/develop/examples/03_Multi-regional_Energy_System_Workflow/InputData/SpatialData) in the Github repository. Would it be possible to also get lp/mps files from them and add them to this PR?

siddharth-krishna:
@JohannesBehrens here are the results of running the latest solvers on the benchmark instances you contributed:
image

Unfortunately, most of them timeout (even with a larger timeout of 24h than we currently use on the platform, which is 10h). Is this consistent with your experience? Could you please share what solver you usually use, how much RAM memory your machine has, and what solver options you usually use? Thank you.

JohannesBehrens:
@siddharth-krishna perhaps a full time series was a little ambitious. I replaced the models with ones that have a lower temporal resolution. I hope the models run now. We usually use them with time series aggregation. If you are still having trouble with the multi-regional model, you can set the MIP Gap to 0.1% or 0.5%.
I have also included two more models from the Energyland example. You can find these under the following link, where a graphical description is also attached: https://github.com/FZJ-IEK3-VSA/FINE/blob/develop/examples/02_EnergyLand/02_Energyland.ipynb

PR #341, opened 2025-11-27
Benchmarks: add ethos.fine.europe model

phil-fzj:
# Add ethos.fine.europe to solver benchmark.

Note that on our machines gurobi (with 8 threads) solves the model within 31000-3500 seconds, which is just below your timeout limit of 10h.
Also note the following gurobi solver settings that we are using:
> BarConvTol=1e-8 method=2 crossover=0 BarHomogeneous=1 ScaleFlag=1

I.e. we are only using barrier.

### Checklist

For PRs adding new benchmark instances:
- [x] I have tested that this model instance can be solved to optimality with gurobi solver [please indicate which solver]
- [ ] (Benchmark team) has tested that some solver solves these benchmarks within our timeouts

### DIFF
--- a/benchmarks/ethos.fine/metadata.yaml
+++ b/benchmarks/ethos.fine/metadata.yaml
@@ -16,7 +16,7 @@ benchmarks:
     Sizes:
     - Name: 175-720ts
       Size: L
-      URL: https://storage.googleapis.com/solver-benchmarks/ethos_fine_europe_60tp-175-720ts.mps
+      URL: https://storage.googleapis.com/solver-benchmarks/instances/ethos_fine_europe_60tp-175-720ts.mps.gz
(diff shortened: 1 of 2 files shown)


siddharth-krishna:
@phil-fzj thank you very much for your benchmark contribution! I've updated the metadata with automatically-computed problem stats, and have launched a run of your benchmarks on our cloud infrastructure. I will share the results here when available, and if it looks good to all of us, we can merge this PR. 🚀

Thanks for sharing the solver settings. As you can see in our Methodology page, for now we run all solvers with default options (and assume the responsibility of picking sensible defaults lies with the solver developers). 

Edit: I've also copied the file from your fileshare to our platform, so you can delete it now if you like.

PR #346, opened 2025-12-01
Website: update dependencies in package.json

jacek-oet:


siddharth-krishna (approved):
Thanks, Jacek. I didn't have time to test it, so please can you test that the website functions as expected before merging it? 🙏

jacek-oet:
@siddharth-krishna 
I tested it, and it works the same as https://openenergybenchmark.org/.

However, I found an issue (not related to this PR):
When a user selects a filter on the Detail Result page and then goes to Home → Key Insights page, an error occurs.
I will fix it today.

image

PR #350, opened 2025-12-04
Update PR template & template metadata

siddharth-krishna:
Updated the PR template based on info we've been asking contributors to provide in recent PRs, and info on benchmark naming conventions. Also updated the template metadata to remove stats field that are computed automatically by `categorize_benchmarks.py`.

Checklist: not applicable

### DIFF
--- a/benchmarks/oemof/metadata.yaml
+++ b/benchmarks/oemof/metadata.yaml
@@ -1,0 +1,297 @@
+benchmarks:
+  OEMOF-diesel-genset-nonconvex-investment:
+    Short description: Hybrid mini-grid model combining Investment and NonConvex options for a diesel generator, PV, battery, and power-conversion components.
+    Modelling framework: OEMOF
+    Model name: diesel_genset_nonconvex_investment
+    Version:
+    Contributor(s)/Source: Vamsi Priya Goli, Open Energy Transition; https://github.com/oemof/oemof-solph/blob/dev/examples/invest_nonconvex_flow_examples/diesel_genset_nonconvex_investment.py
+    License: CC BY 4.0
+    Problem class: MILP
+    Application: Infrastructure & Capacity expansion
+    Sectoral focus: Power-only
+    Sectors: Electric
+    Time horizon: Single-stage
+    MILP features: NonConvex flows, investment optimization, binary on/off for genset
+    Sizes:
+    - Name: 1-240ts
+      Size: S
+      URL: https://storage.googleapis.com/solver-benchmarks/instances/OEMOF-diesel-genset-nonconvex-investment-1-240ts.lp.gz
+      Temporal resolution: 240 time slices
+      Spatial resolution: Single-node mini-grid
+      Realistic: false
+      Num. constraints: 4331
+      Num. variables: 3374
+      Num. continuous variables: 3134
+      Num. integer variables: 240
+  OEMOF-economic-results:
+    Short description: Simple energy system model demonstrating variable and investment cost calculation using oemof.solph Results class.
+    Modelling framework: OEMOF
+    Model name: economic_results_with_invest
+    Version:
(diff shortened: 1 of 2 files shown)


benchmarks/_template_metadata.yaml
```diff
@@ -1,6 +1,6 @@
 ## (optional) link to the modelling framework repo used to generate the benchmark(s) (e.g. https://github.com/PyPSA/pypsa-eur)
 benchmarks:
-  Benchmark_name: # Add the benchmark name
+  Benchmark_name: # Add the benchmark name (convention: [model-name]-[country/region]-[application:trex/op/etc])
```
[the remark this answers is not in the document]
danielelerede-oet: Yes I'd say that;s fine, I would only add a comment to highlight that we don't want spatial/time resolution here, but in the 'Name' field

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,15 +1,31 @@
...
 
-For PRs adding new benchmark instances:
+<!-- Delete the section(s) that are not relevant to this PR type -->
+
+**For PRs adding new benchmark instances:**
 - [ ] I consent to releasing these benchmark instance files under the CC BY 4.0 license
-- [ ] I have tested that this model instance can be solved to optimality with {solver} solver [please indicate which solver]
-- [ ] (Benchmark team) has tested that some solver solves these benchmarks within our timeouts
+- [ ] The benchmark name and size instance name follow the conventions indicated in the template
+- [ ] I have tested that this model instance can be solved to optimality with [solver] solver, using [options], on a [spec] machine <!-- please indicate which solver, how much RAM memory your machine has, and whether you used any non-default solver options -->
```
danielelerede-oet: This is definitely the most important field. For solver options, I'm scared that the default may vary by modelling framework / configuration within the same modelling framework, so we should maybe ask to be more explicit?
Also, an indication about the runtime obtained with the specified option would guarantee that the test was actually performed

benchmarks/_template_metadata.yaml
```diff
@@ -15,11 +15,8 @@ benchmarks:
     MILP features: # Current set (feel free to propose new ones if necessary): None, Unit commitment, Piecewise fuel usage, Transmission switching, Modularity
     Sizes:
     - Name: # Spatial resolution-Temporal resolution, e.g., 2-12ts, 10-1h
-      Size: # Automatically assigned when running the benchmark
```
danielelerede-oet: What's the reason to remove this field? Is it compiled automatically later?

benchmarks/_template_metadata.yaml
```diff
@@ -15,11 +15,8 @@ benchmarks:
...
       URL: # link to the shared folder where the .lp / .mps file is stored, maintainers will assign a new link after accepting the contribution
       Temporal resolution: # n. of time steps in one planning horizon, e.g., 1 hour / 8 time slices
       Spatial resolution: # n. of nodes / regions, e.g., 997 nodes / 25 regions
       Realistic: # true, false
-      Realistic motivation: # (optional) e.g., Sufficient spatial and temporal resolution over the considered time horizon to allow the benchmark to be defined as Realistic.
-      Num. constraints: # Automatically assigned when running the benchmark
-      Num. variables: # Automatically assigned when running the benchmark
```
danielelerede-oet: Same as comment above, so I guess I know the answer

benchmarks/_template_metadata.yaml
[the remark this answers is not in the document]
danielelerede-oet: I'm not sure if you checked the other comments @siddharth-krishna

benchmarks/_template_metadata.yaml
siddharth-krishna: Yes, there has been a few instances now where the size reported by people was incorrect, so I thought it simpler if we add it using the `categorize_benchmarks.py` script :)

PR #356, opened 2025-12-14
Benchmarks: add SWITCH-China model

Vamsipriya22:
This PR contains benchmark function for the SWITCH-China model.

### Checklist

For Benchmark team:
- [x] Run `tests/validate_urls.py` to ensure URLs are consistent with benchmark and size instance name
- [x] Test that some solver solves these benchmarks within our timeouts on our infra

### DIFF
--- a/benchmarks/fine/metadata.yaml
+++ b/benchmarks/fine/metadata.yaml
@@ -1,0 +1,207 @@
+benchmarks:
+  FINE-2-node-electricity-supply-system:
+    Short description: A simple 2-region energy supply system built using the ETHOS.FINE framework, illustrating the creation of an energy system model and the addition of sources, converters, storage, transmission, and sinks.
+    Modelling framework: FINE
+    Model name: 2_node_electricity_supply_system
+    Version:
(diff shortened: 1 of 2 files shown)


benchmarks/switch/metadata.yaml
```diff
@@ -1,4 +1,29 @@
 benchmarks:
+  SWITCH-China-open-model:
+    Short description: A large-scale national energy system model of China built using the SWITCH framework, including extended China-specific modules and updated inputs based on a 2020 base year.
+    Modelling framework: SWITCH
+    Model name: SWITCH_China_open_model
+    Version:
+    Contributor(s)/Source: Vamsi Priya Goli, Open Energy Transition; https://github.com/switch-model/switch-china-open-model
+    License: CC BY 4.0
+    Problem class: LP
```
siddharth-krishna: Are we sure it is a MILP? The categorize script found 0 int vars, see below.

danielelerede-oet: Hey Sid I think it's an LP, but that's also what I see in the metadata

PR #377, opened 2026-01-20
Benchmarks: Add Spineopt instances

datejada:
This pull request adds a new benchmark entry for the SpineOpt modeling framework to the `benchmarks/spineopt/metadata.yaml` file. The new entry provides detailed metadata for the "spineopt-eu-pathway-planning" benchmark, including its description, contributors, license, problem class, application area, sectoral focus, time horizon, and dataset characteristics.

Also, adding a README file for indications on the solver options to handle numerical issues when running this file. For example, running without crossover and passing specific scaling options to HiGHS. 

### Checklist

- [x] I have tested that this model instance can be solved to optimality in 4h with **HiGHS v1.20.0 [755a8e027]** solver, using options: "solver" to "hipo", "parallel" to "on", "threads" to 48, "run_crossover" to "off", "user_objective_scale" to -7, "user_bound_scale" to -12, on a machine with 392913 MB, 64 Cores 3.6 GHz, and **OS Windows**

Log: [log-20260113b.txt](https://github.com/user-attachments/files/26442829/log-20260113b.txt)

EDIT: Add HiGHs version, OS, and log info of the run.


datejada:
> > > Dear @datejada thank you for your contribution! I have made a few changes in the metadata and posted a few comments on a few fields. I also have two questions for you:
> 
> > > 
> 
> > > * As you may know, we're currently only using default solver options for every solver we include in the platform and a VM with 62 GB memory and 8 vCPUs. Your choices look very specific though, especially concerning the solver (any option for that, did you test anything else?). Do you also have an idea about how the runtime scales if default HiGHS options are used (BTW we'll be also including HiPO and other solver options for other solvers in the future)?
> 
> > > * Do you have the possibility to easily scale down the model to a represent, for instance, a single country or a more limited set of countries? And is it possible to add any MILP features (e.g. unit commitment for conventional generators) to get more instances based on this problem?
> 
> > 
> 
> > Dear @danielelerede-oet,
> 
> > 
> 
> > Thank you for your review and questions. I hope my updates are now aligned better with the benchmarks. If there are still remaining issues, please let me know, and I will be happy to update them accordingly. Regarding your questions:
> 
> > 
> 
> > * The VM specification should be enough to solve this problem. It would take longer than the times I got, but I saw peak RAM use around 60GB when using HiGHS. So, it should be okay. Same with the threads, 8 should be okay to run this case. Regarding the other options ("solver" to "hipo", "parallel" to "on", "run_crossover" to "off", "user_objective_scale" to -7, and "user_bound_scale" to -12) are a must, otherwise HiGHS can't solve that problem. The only solver option that worked with those problems was HiPO. In addition, the crossover must go off (even for commercial solvers like Gurobi), and the scaling factors are necessary because this problem has scaling issues; those values help HiGHS solve the problem. In short, I think this type of requirement (i.e., passing solver options) will be more common for real case studies, like this one (and even for commercial solvers like Gurobi). I can create an issue for it, if you think that would help to push the need internally 😉 (a priori, I assume it is easy to pass that information to the solver to the VM, but I don't know the details of the implementation)
> 
> > * It is possible to scale down the problem in the temporal and in the number of nodes. We have workflows that help create those instances if needed (we plan to add more instances to this repo in the future). And yes, it is possible to add MILP features, like unit commitment or integer investment decision, but with this size, I hardly doubt that we can solve it as an MILP problem 😅 maybe for smaller instances, that is a possibility.
> 
> 
> 
> Thanks @datejada!
> 
> 
> 
> - Regarding your choice for the solver I understand it perfectly, I was just curious about how you came up with those specifications. I've also run it in Gurobi (v 13.0.0) on the same VM we use for running the website and (without crossover) it took around 10 hours. You're more than welcome to open an issue with any other additional suggestion on  solver options. We'll be working on that in the coming months BTW, but getting feedback is also very important to us.
> 
> - Regarding scaling down and adding MILP features, we'd be more than happy to get more about them, and even large (maybe not XL) MILP problems as we're currently lacking of them.
> 
> 
> 
> In addition, I have a question on the representative days to understand how to specify that in the metadata without massive changes to the current structure of the website. I guess you run the model for a full year but considering the different represnetative days for  renewable availability, different demand profiles etc., and do you also consider a hourly resolution within each day?

@danielelerede-oet thanks for the reply. It is great that you are working on adding the options for the solvers. I will be happy to contribute with more cases once they are ready 😊
As for the representative days, yes they are with hourly resolution. So, I think it is fine leaving it as hours, since the name has the representative days and the number there.
Thanks! And please let me know if you have any other comments or questions

danielelerede-oet (approved):

PR #381, opened 2026-01-25
Benchmarks: add SWITCH-USA-PG model with reduced number of load zones

Vamsipriya22:
This PR contains benchmark function for the SWITCH-USA-PG model with the reduced number of load zones.

### Checklist

For Benchmark team:
- [ ] Upload the LP/MPS files (compressed using `gzip -9 <filename>`) to our GCS bucket
- [ ] Run `benchmarks/categorize_benchmarks.py` on them to obtain problem stats and size category


Vamsipriya22:
> Thanks @Vamsipriya22 , the same comments from #382 apply here

I’ve updated it to use a single SWITCH-USA-PG model entry with instances under "Sizes".

Please review it when you get a chance.

danielelerede-oet (approved):

PR #382, opened 2026-01-25
Benchmarks: add SWITCH-China model with reduced number of load zones

Vamsipriya22:
This PR contains benchmark function for the SWITCH-China model with the reduced number of load zones.

### Checklist

For Benchmark team:
- [ ] Upload the LP/MPS files (compressed using `gzip -9 <filename>`) to our GCS bucket
- [ ] Run `benchmarks/categorize_benchmarks.py` on them to obtain problem stats and size category


danielelerede-oet:
> Thank you for the review.
> 
> Using a single model with different sizes helps isolate the effect of problem size on solver performance while keeping the model structure constant. By keeping the same base model (SWITCH) and varying only the number of load zones, runtime differences mainly reflect scaling (i.e., an increase in variables and constraints), rather than changes in formulation logic across models. This makes the comparison cleaner and more controlled for benchmarking purposes.
> 
> A README has also been added as suggested, explaining how the different sizes are generated and how the benchmarks are executed.

Sorry @Vamsipriya22 my question wasn't really clear. I meant why don't we just have SWITCH-China-open-model and then the different instances under "Sizes" as we do, for instance, with PyPSA benchmarks?

Vamsipriya22:
@danielelerede-oet  I’ve updated it accordingly.

Now we have a single `SWITCH-China-open-model` entry with the different instances organized under **"Sizes"**, similar to how it’s structured for the PyPSA benchmarks.

Please review it when you get a chance and let me know if anything needs further adjustment.

danielelerede-oet (approved):
Wonderful @Vamsipriya22 , I've categorized the benchmarks and merged the metadata and all the checks are now working! Merging!

PR #400, opened 2026-02-02
Website: split key-insights main plot into 3: LP/MILP/HiPO

### LINKED ISSUE #398: Website: split main results plot into LP/MILP in key-insights page
We want to split the overall results plot on the key-insights page (of the `run-v2` branch!) into 2 plots, one for the LP problems and one for the MILPs:
Image
The plots should have numbers that match those in these reference plots:
Image
Image

We also want to add a 3rd plot to this section, which shows the LP results and includes the HiPO and IPM solvers. These results are in `results/benchmark_results_hipo.csv` of this branch:
Image

Finally, it would be good if we could add the % of solved problems in light grey to the top of the bars, as is done in the reference plots above.

jacek-oet:
Fixes #398

image

image

image


siddharth-krishna (approved):
Looks good thank you!

Also, is this order of categories in the x-axis fixed in your code? if so / if it's easy to change it, can we please use the order: small, medium & realistic, medium, large & realistic, large?
Image

PR #401, opened 2026-02-02
Website: Review Methodology section

danielelerede-oet:
This PR reviews the Methodology sector for v2 (February 2026)

### Benchmark contributions

Please add the following information to the description field of the metadata for each benchmark instance:
- Model code link
- Any paper describing the energy system model and its equations
- Data sources used in constructing the model
- Reference to any real-world study or published paper that this model was used for

Please add a README with instructions on how to run the model framework's code to reproduce benchmark generation, particularly if the code doesn't automatically generate MPS or LP files (see, for instance, this [example](https://github.com/open-energy-transition/solver-benchmark/blob/main/benchmarks/switch/README.md)).

### Checklist

### DIFF
--- a/website-nextjs/src/components/methodology/MethodologySection.tsx
+++ b/website-nextjs/src/components/methodology/MethodologySection.tsx
@@ -164,9 +164,9 @@ const MethodologySection = () => {
               </li>
               <li className="mb-2">
                 We do not average over multiple random seeds for now, in order
-                to save time, costs and environmental impact. This is
-                something we may consider in the future if there is interest and
-                budget for it.
+                to save time, costs and environmental impact. This is something
+                we may consider in the future if there is interest and budget
+                for it.
               </li>
             </ul>
           </li>


website-nextjs/src/components/methodology/Metrics.tsx
```diff
@@ -74,7 +74,7 @@ const Metrics = () => {
       <p>
         After running benchmarks, we manually check any runs where the above 2
         metrics are above <code>1e-4</code> for errors. In our results so far,
-        no solver had a max integrality violation of above <code>1e-5</code>.
+        no solver had a max integrality violation of above <code>1e-7</code>.
```
siddharth-krishna: Actually, the results on the run-v2 branch does have some instances above 1e-7:
Image

But hopefully 1e-5 is acceptable?

danielelerede-oet:
> Thanks, Daniele for catching these updates! If you don't mind, I'll close this PR and add the diff to the `run-v2` branch in commit https://github.com/open-energy-transition/solver-benchmark/commit/6df27f9c7c124a9cd1f6322596236d6f2712d658 and https://github.com/open-energy-transition/solver-benchmark/commit/65aff12d32fd4ace790fa01f02934755e2cc9a5d
> 
> Then we can update the website in one go with new results and new text reflecting the new timeout etc.

Hi @siddharth-krishna that's fine. I opened a new PR to make changes more visible but I agree with having the changes in the run-v2 PR. What about the point on runtime error estimation though?

PR #416, opened 2026-02-09
Runner: per-solver conda envs with Docker support (#411)

### LINKED ISSUE #411: Runner: use one conda env per solver

HeathNaylor:
This PR adds conda environments for each solver so that the solver and version pair runs in it's own environment. I created a new `runner/solvers.yaml` registry to map solvers to their version, year, and env. `run_benchmarks.py` uses `conda run` to switch envs automatically.

I'm doing this work on my macbook so I needed a Linux environment per the documentation. I figured I'm not the only one doing this so I decided to add a `runner/Dockerfile` based on continuumio/miniconda3 so we can run these benchmarks in a Linux container. I've made it optional in the docs. If the Docker work is out of scope for this I can remove it from the PR.

One thing worth noting is that I added a check for systemd so that if you don't have it then memory limits are just skipped (and the user notified in the command line) so theoretically you could run this on MacOS or Windows too. This could also be removed from the PR if desired.

Other changes:
- benchmark_all.sh continues past conda env create failures instead of aborting the whole run. I was finding licensing issues for Gurobi and missing packages from PyPI.
- Fixed a bug where log_file.exists (missing parens) was always truthy, causing crashes when log files didn't exist.

Missing package error:
```
ERROR: Could not find a version that satisfies the requirement highspy==1.5.0.dev0
  (from versions: 1.7.1.dev1, 1.7.1, 1.7.2, 1.9.0, 1.10.0, 1.11.0, 1.12.0, 1.13.0)
  ERROR: No matching distribution found for highspy==1.5.0.dev0
```

Gurobi license errors:
```
  Gurobi 10.0.0 (2022): gurobipy.GurobiError: License expired 2024-10-28
  Gurobi 11.0.0 (2023): gurobipy.GurobiError: License expired 2025-11-24
```

Fixes #411


eantonini (approved):
@HeathNaylor thanks a lot for your contribution. This is a solid step forward for isolating solver environments, and the log_file.exists() fix is a legit bug catch. Thanks also for flagging the Docker/systemd scope questions explicitly; I think both are worth keeping.

We are about to do a broader refactor of the runner, which will likely touch this area again, so I am approving it now rather than blocking on cleanup. Tracking the following as follow-ups for that refactor rather than asking for changes here:

- Dead code: get_conda_package_versions() in run_benchmarks.py is no longer called now that get_solver_versions_from_config() replaces it — can be removed.
- Orphaned env files: the old per-year files (runner/envs/benchmark-2020.yaml through benchmark-2025-fixed.yaml) are no longer referenced by benchmark_all.sh's non-test path (only benchmark-tests[-fixed].yaml still is) and should be removed once we're confident nothing external depends on them.
- prefix: vs name: mismatch in generated fixed YAMLs: several -fixed.yaml files (cbc-2023/2024, glpk-2020, gurobi-2022–2025, highs-2025) carry a stale prefix: line (e.g. solver-cbc-2.10.11) that doesn't match the name: field (benchmark-cbc-2023) the rest of the tooling relies on. Modern conda ignores prefix: on env create, so this shouldn't break anything, but generate_fixed_envs.sh should strip it to avoid confusion.
- Uneven "fixed" coverage: highs-2022/23/24 and all scip years don't have -fixed.yaml counterparts, so they always install from the loose spec. Fine functionally, but worth deciding whether we want full pinning across the board.
- Silent partial failures: benchmark_all.sh now continues past a failed conda env create with just a WARNING. Intentional per the PR description (Gurobi license/PyPI issues), but means a solver can silently be missing from a run's results. It is worth deciding if that should be surfaced more loudly (e.g. summary at the end of the run).

None of these should affect the correctness of this PR.

@danielelerede-oet @siddharth-krishna anything else that should be flagged?

Issue #82, opened 2024-12-12
Getting PyPSA benchmark instances


jajhall:
I've tried following the instructions on 

https://github.com/open-energy-transition/solver-benchmark/blob/main/benchmarks/pypsa/README.md

but get the following error

![image](https://github.com/user-attachments/assets/b9f7b702-2392-4747-9915-267b34b07f8d)

Note that, although I didn't reach the line

> docker run -it --entrypoint bash <image-hash-from-previous-output>

I'm puzzled as to what `<image-hash-from-previous-output>` refers.

If relevant, I tried following 

https://github.com/open-energy-transition/solver-benchmark/tree/main?tab=readme-ov-file#initial-setup

but 

> sudo pip install -r website/requirements.txt

was not possible since I'm on an externally managed desktop

Since I'm not intending to run benchmarks using your "system", or use the website locally, it's probably not worth trying to get ting the above to work for me, but I hope that it's useful feedback from a user

Please make the PyPSA benchmark instances available as a folder of .lp files (as we do [here](https://github.com/jump-dev/open-energy-modeling-benchmarks/tree/main/instances))  and/or let me know where I can download them otherwise.

siddharth-krishna:
https://github.com/open-energy-transition/solver-benchmark/pull/328 introduced a bulk download UI to the benchmark set [page](https://openenergybenchmark.org/dashboard/benchmark-set) of the website -- so hopefully it is easy to download LP/MPS files for any desired subset of benchmarks.

I'm closing this issue, improving the generation scripts for the PyPSA instances is being done in https://github.com/open-energy-transition/solver-benchmark/pull/312

PR #424, opened 2026-02-11
Benchmark: re-work machinery for generating the pypsa benchmarks

finozzifa:
This pull request adds:
- Introduce a separate `benchmarks/plain_pypsa` sub-directory that contains the plain `pypsa` generating scripts
- Remove `generate.sh` and replace it with `generate.py`
- Update the `Dockerfile` to use `pixi`
- Update the `README.md`
- Introduced and updated the benchmark specific configurations
- Updated the `solve_network.patch`

The issue was prompted by Issue #82.


finozzifa:
oki I am on it. Thanks for clarifying it

PR #426, opened 2026-02-12
Benchmarks: Add PyPSA-Eur and PyPSA-DE MILPs

danielelerede-oet:
### Benchmark contributions
This PR adds PyPSA-Eur (50,100 nodes, 1h-3h,24h,168h temporal resolution) and PyPSA-DE (10,20,50,100 nodes, 1h-3h,24h,168h temporal resolution) MILP instances considering non-linearized Unit Commitment and generated using the PyPSA benchmark generator in #424 .

No need to update the summary table here.

Please add the following information to the description field of the metadata for each benchmark instance:
- Model code link https://github.com/PyPSA/pypsa-eur (commit 665315c)

### Checklist

For Benchmark team:
- [x] Run `tests/validate_urls.py` to ensure URLs are consistent with benchmark and size instance name
- [x] Add CC BY 4.0 license to the metadata file


siddharth-krishna (commented):
Thanks for this, Daniele! I'm thinking perhaps we wait for the current experiments with the MILP spatial scaling study to see what instances are solvable in our limits before merging this in? Perhaps we don't want to add so many instances if most of them are too large

Btw I also notice this PR checks in some files to the `website-nextjs/` folder. They should be removed from this branch. A recent PR moved our website to `website/`, so I think that has caused the issue. Thanks!

danielelerede-oet:
@siddharth-krishna @eantonini this should be ready to go!

siddharth-krishna (changes requested):
Thanks, Daniele! A few requests please:
- Could you pelase remove the changes to `website-nextjs/` from this PR branch? I think those are leftovers because we moved the website to `website/` during the time this PR was open
- Could you please remind me (and maybe update this PR description) why we are removing some instances (e.g. `pypsa-de-elec-9-1h`)?
- The added instances (e.g. `pypsa-eur-elec-trex_copt-uc-dfp-50-3h`) do not have results in `results/benchmark_results.csv`. I assume this is because they're not used for v3 website results but rather our scaling study for the report? Please also update the PR description with this because my memory is bad hahaa 🙏

danielelerede-oet:
> Thanks, Daniele! A few requests please:
> 
> * Could you pelase remove the changes to `website-nextjs/` from this PR branch? I think those are leftovers because we moved the website to `website/` during the time this PR was open
> * Could you please remind me (and maybe update this PR description) why we are removing some instances (e.g. `pypsa-de-elec-9-1h`)?
> * The added instances (e.g. `pypsa-eur-elec-trex_copt-uc-dfp-50-3h`) do not have results in `results/benchmark_results.csv`. I assume this is because they're not used for v3 website results but rather our scaling study for the report? Please also update the PR description with this because my memory is bad hahaa 🙏

Thanks Sid, I also noticed some other changes to be done and I've pushed them. Regarding the removal of a few instances like `pypsa-de-elec-9-1h` I did it because for the State of solvers report / HiPO blogpost we only used instances with an even n. of nodes. So, I wanted to reflect that also because if we have a 8 nodes and 10 nodes version of the same model, having the 9 nodes version doesn't add too much value at the cost of much higher time/money expenditure to run all of them (the same applies to the others I removed).

Regarding the PyPSA-Eur versions I was also in doubt and wanted to check this with you. We haven't used them for the scaling studies but would be a *nice to have* for me. Would you prefer me to test them against TO/OOM before adding them to the repo?

siddharth-krishna (approved):
Thanks, Daniele! Good idea to remove the instances with odd number of nodes that we didn't use.

> Regarding the PyPSA-Eur versions I was also in doubt and wanted to check this with you. We haven't used them for the scaling studies but would be a nice to have for me. Would you prefer me to test them against TO/OOM before adding them to the repo?

Happy to have them! Perhaps we can add a comment in the metadata file explaining that these haven't been used for website or scaling study but can be used for v3, so that we remember?

PR #433, opened 2026-02-19
Runner: add CPLEX, Knitro, Xpress, and Mosek solver support

KristijanFaust-OET:
Implements CPLEX, knitro and xpress commercial solver runs.

### Knitro
Knitro support confirmed the following options
```
feastol=1e-6
opttol=1e-6
mip_opt_gap_rel=1e-4
```
- [ ] how to say we don't need a basic solution?

### Xpress

- [ ] Have we heard back from FICO support?

### Gurobi

Gurobi team recommended
```
      MIPGap: 1.e-4 # default
      FeasibilityTol: 1.e-6 # default
      OptimalityTol: 1.e-6  # default
      SolutionTarget: 1  # basic not needed
```
They don't think it's worth trying to set `BarConvTol` as the meaning of this varies too much between solvers.

### Cplex

Cplex support confirmed
```
            "mip.tolerances.mipgap": mip_gap,
            # LP tolerances
            "simplex.tolerances.feasibility": lp_tol,
            "simplex.tolerances.optimality": lp_tol,
            "solutiontype": 2, # Basic solutions not needed
```
- [ ] Have asked: do the simplex tolerances also work for barrier?


danielelerede-oet:
@KristijanFaust-OET @siddharth-krishna I've added Mosek to the set of commercial solvers again considering default options only.

In addition, I see you're using `solver_model.get_time()` for CPLEX, but that metric does not provide a clean solver runtime comparable to other solvers. It reflects an internal clock rather than the time spent solving a single instance, and may include additional overhead or accumulated time. As far as I know, CPLEX does not expose a directly comparable runtime metric through this interface. There is some ongoing work in Linopy to improve this (https://github.com/PyPSA/linopy/pull/583) while I also provided a workaround that we could use here (https://github.com/PyPSA/linopy/pull/636).

infrastructure/startup-script.sh
```diff
@@ -43,9 +43,65 @@ chmod a+x /usr/local/bin/yq
 mkdir -p /opt/gurobi
 gsutil cp gs://solver-benchmarks-restricted/gurobi-benchmark-40-session.lic /opt/gurobi/gurobi.lic
 
+# Download MOSEK license
+echo "Downloading MOSEK license..."
+mkdir -p /opt/mosek
+gsutil cp gs://solver-benchmarks-restricted/mosek_evaluation_unrestricted.lic /opt/mosek/mosek.lic
```
danielelerede-oet: @siddharth-krishna do you confirm that it's safe to store this kind of information here?

infrastructure/main.tf
```diff
@@ -56,7 +56,7 @@ variable "enable_gcs_upload" {
 variable "gcs_bucket_name" {
   description = "Name of the GCS bucket to upload benchmark results"
   type        = string
-  default     = "solver-benchmarks"
+  default     = "solver-benchmarks-testing"
```
siddharth-krishna: This should also be reverted

infrastructure/startup-script.sh
```diff
@@ -43,9 +43,65 @@ chmod a+x /usr/local/bin/yq
...
+    echo "CPLEX installation completed successfully"
+else
+    echo "ERROR: CPLEX binary not found - installation failed"
+    ls -la /opt/ibm/ILOG/CPLEX_Studio2212/ 2>/dev/null || echo "CPLEX install directory does not exist"
+fi
+
 # Clone the repository
 echo "Cloning repository..."
-git clone --depth=1 -b main https://github.com/open-energy-transition/solver-benchmark.git
+git clone --depth=1 -b commercial-solvers-run https://github.com/open-energy-transition/solver-benchmark.git
```
siddharth-krishna: This should be reverted too

infrastructure/startup-script.sh
siddharth-krishna: It was fine for an ad-hoc experiment, but we should eventually implement something more robust: e.g. passing a URL as a parameter to tofu

infrastructure/startup-script.sh
```diff
@@ -264,12 +327,7 @@ if [ "${ENABLE_GCS_UPLOAD}" == "true" ]; then
         echo "Uploading ${compressed_file} to GCS bucket..."
 
         # Check if file contains "gurobi" in the name
-        if [[ "${filename}" == *"gurobi"* ]]; then
-            echo "File contains 'gurobi' in name, storing in restricted folder..."
-            gsutil cp "${compressed_file}" "gs://${GCS_BUCKET_NAME}-restricted/logs/${RUN_ID}/${filename}.gz"
-        else
-            gsutil cp "${compressed_file}" "gs://${GCS_BUCKET_NAME}/logs/${RUN_ID}/${filename}.gz"
-        fi
+        gsutil cp "${compressed_file}" "gs://${GCS_BUCKET_NAME}/logs/${RUN_ID}/${filename}.gz"
```
siddharth-krishna: This should be reverted too

runner/run_solver.py
```diff
@@ -62,33 +63,77 @@ def get_solver(solver_name):
...
             "randomCbcSeed": 1,  # 0 indicates time of day
             "ratioGap": mip_gap,
         },
+        # TODO also set BarConvTol for all solvers?
+        "highs": {
+            "random_seed": 4,
+            "mip_rel_gap": mip_gap,
+            # LP tolerances
+            "primal_feasibility_tolerance": lp_tol,
+            "dual_feasibility_tolerance": lp_tol,
```
siddharth-krishna, line 82: This PR set some LP tolerances in order to make more fair comparisons. We should consider doing this also for v3 (perhaps there's already an issue open about this? I can also dig through my emails for what each solver developer recommended.)

siddharth-krishna (commented):
Enrico, most of this PR has changes for a specific ad-hoc experiment and should be reverted, I left a few comments. Thanks!

eantonini (approved):
Thanks both for the review, addressing each thread:

Rregarding @siddharth-krishna's four infra revert requests, they are all done, and taken further:
- `infrastructure/main.tf`'s `gcs_bucket_name` default (`solver-benchmarks-testing` → `solver-benchmarks`) is reverted.
- `infrastructure/startup-script.sh`'s `git clone -b commercial-solvers-run` is back to `-b main`.
- The MOSEK/Knitro/Xpress license downloads + CPLEX silent-install block are removed entirely.
- The restricted-bucket log routing is back to the original gurobi-only check.

`infrastructure/startup-script.sh` and `infrastructure/main.tf` are now identical to `main`. This also resolves @danielelerede-oet's original "is it safe to store this here?" question by removing the code rather than answering it. Provisioning CPLEX/MOSEK/Knitro/Xpress licenses on GCE VMs (ideally via something more robust than a committed script) is being pushed to a dedicated follow-up infra PR, so it gets its own focused security review. The Python-side solver support (`run_solver.py`, `solvers.yaml`, the per-solver env files) is untouched. CPLEX/MOSEK/Knitro/Xpress remain usable if you bring your own license/environment.

@siddharth-krishna on LP tolerances (`run_solver.py` line 77): yes, intentional and kept. We will find a more robust way of defining them in a config file during the refactoring.

@danielelerede-oet on CPLEX runtime (`get_reported_runtime`, still `solver_model.get_time()`): I believe this was fixed upstream on linopy. Can you confirm?

Everything above is on top of the earlier round of fixes.

danielelerede-oet:
@eantonini I checked it again and it's good to go for me

PR #438, opened 2026-02-23
Results: re-run high variance results, update website

siddharth-krishna:
This PR re-runs instances with variance > 6%, and updates the results and website accordingly. It also:
- Updates the util function to create benchmark campaign to allow setting years
- Fixes logging for hipo and ipx in cases where it TOs
- Updates validate_results.py to ignore benchmarks marked `Skip_because`
- Removes benchmarks we decided not to run
- Updates the website with latest VM type and variance threshold used in v2

Before merge:
- [x] When the last instance finishes running, update results CSV

### Checklist


eantonini (approved):
Look good, thanks.

PR #441, opened 2026-02-24
Website: explain different timelimits on compare-solvers; re-order filters

### LINKED ISSUE #217: Timelimits unclear
There appear to be instances that no solver can solve within the time limit. But there are other instances or benchmarks that have been given a larger time limit:

Image

This just begs the question of why such a "short" time limit was set, which apparently is not sufficient for any solver.

siddharth-krishna:
Fixes #217 

Also re-orders the filters on the left of all dashboards to have LP/MILP, Size, Realistic at the top -- because those are the ones I use most often! But please disagree if you do otherwise, it's easy to re-order.

image

### Checklist


jacek-oet (approved):
LGTM

PR #463, opened 2026-03-03
Website: add arrows lower is better help interpret plots

### LINKED ISSUE #422: Website: add arrows "lower is better" etc. to help interpret plots

jacek-oet:


eantonini (approved):
Everything looks good to me, thanks!

eantonini (changes requested):
Thanks for implementing these changes.

I think they all look great except the "compare solvers" dashboard. Here I would try to implement the arrows as I sketched above, because the solution that you implemented is not very clear.

When you add these arrows and text to the "compare solvers" dashboard, try to plot the data points on top of these if there is some overlap (i.e., the first layer has the data points, the second layer the arrows and text). Let me know if this is not clear.

eantonini (changes requested):
Thanks for implementing this. The arrows and text look very good and increase the readability of the plot a lot.

I have one further request regarding the colors of these elements. In my previous sketch I used blue and orange for the arrows and text thinking to the colors that the two solvers have on the other dashboard. I didn't notice that in this dashboard orange and blue refer to the crosses indicating that either or both of the solvers fail.

Could we match the colors that we have for the solvers on the other dashboards to the colors that we have here and find other colors for the crosses?


I also noticed two bugs: the grid extends outside the plot area of the left plot and one x-axis tick label is missing from the right plot.

Image

eantonini:
@jacek-oet Thanks for implementing my last requests and fixing the bugs.

I have two additional requests:
- make the text ("cbs is better", etc) bold,
- bring the text and arrows to the front of the data points (on top of them), not to the back.
image

jacek-oet:
@eantonini Please check.
image

eantonini (approved):
Looks great, thanks!

PR #465, opened 2026-03-05
Improve CI checks - Issue #419

### LINKED ISSUE #419: Improve CI checks
- [ ] Ensure metadata URLs don't use `storage.cloud.google.com`, but use `storage.googleapis.com` as the latter doesn't need sign-in
- [ ] Check all metadata files have the same yaml keys (is there a way to define a schema)
- [ ] `merge_metadata.py`'s warnings/errors are silent right now, they don't fail CI
- [ ] Ensure unique bench names, size names, URLs
- [ ] Run many S benchs on all solvers on CI to avoid linopy/solver bugs?

Vamsipriya22:
This PR is related to the issue https://github.com/open-energy-transition/solver-benchmark/issues/419.

This PR improves the CI and metadata validation. It addresses the following issues:

✅ Normalize metadata URLs to use storage.googleapis.com instead of storage.cloud.google.com, avoiding sign-in requirements.

✅ Added a schema file to ensure all metadata files follow a consistent set of YAML keys.

✅ Ensure merge_metadata.py warnings and errors fail the CI instead of being silent.

✅ Enforce uniqueness of benchmark names, size names, and URLs across all metadata.

✅ Run multiple sample benchmarks on all solvers in CI to detect potential linopy or solver issues early.


eantonini (approved):
Looks good to me. Thanks for implementing these features!

PR #473, opened 2026-03-20
Website improve solvers dashboard plot

jacek-oet:
- [x] Add text "(click a solver to toggle showing it on the plot)" to the line containing the solvers legend 
- [x] Make the "base solver solved successfully" circle and X on the top right the same the color as the solver
    
- [x] Add arrows on both sides of the left y-axis. Arrow up should say "cbc better than other solver", down should say "other solver better than cbc"
    
- [x] When selecting hipo solver on this dashboard, I think we are also plotting the MILP benchmarks -- but we only ran hipo on LPs. Can we first filter the benchmarks to the set where the selected solver was run?
image


eantonini (changes requested):
The arrows look much better now.

Regarding the plot, is it possible to show by default only two solvers and not all of them? Then allow users to select more if they want.

Besides that, I found a bug. Below are the step to reproduce it:
1. Select only one solver, for example, cbc like in the figure below.
Image
2. Now change the solver to something else, for example gurobi.
Image
3. Here I cannot select and display gurobi's runtime (y-axis on the right).

eantonini (approved):
Looks great, thanks!

PR #475, opened 2026-03-22
Runner: Add a quickstart notebook for benchmark running

siddharth-krishna:
This PR adds a notebook that I used to run a recently submitted set of benchmarks. I tried to document the process as carefully as I could, in the hope that this notebook can be used as a template for others to run benchmarks on the cloud, validate the results, and add them to the website. Questions and suggestions welcome!

Please go through it, and copy this notebook's structure and follow its checklist (learnt the hard way!) whenever you are running benchmarks on the cloud.

### Checklist

N/A


danielelerede-oet:
@siddharth-krishna  I finally had time to review this. It matches my experience in the last weeks in launching benchmark campaigns and so far this is definitely the best option we have.

I have added  a few changes directly in the notebooks and I have a few comments here:

1. When you mention to "Estimate cost before launching a run, it's easy to launch something costing $1000!" what about providing a rough information about the cost per hour for the VM used in the notebook?
2. Do we want to give the possibility to everybody to write results to the website?

Do you see automatizing this process via CLI as a positive improvement to replace the notebook? I think be should be very careful especially considering costs, but it could be a groundbreaking improvement. We have discussed this with @eantonini and I'd like to work on it but I'd also like to hear your thoughts about it.

danielelerede-oet (approved):

PR #481, opened 2026-04-03
Review benchmark nomenclature and update website UI

danielelerede-oet:
The goal of this PR is to update the nomenclature in the metadata schema. This involved two things:
- **Richer, more precise fields**: The new schema adds fields for temporal and spatial scope (e.g. geographic scope, temporal resolution/sampling, number of optimization periods) and other attributes more specific to energy system models. The old fields are kept for now. A follow-up PR will populate the new fields, inferring values from the existing data where possible.
- **Flattened structure**: Each benchmark model+size combination is now a single, standalone problem with its own Problem ID, rather than a benchmark model with nested size variants.

**Website changes**
This PR touches a large number of website files, but the actual changes to the site are minimal. Most of them are small, mechanical follow-ups to the schema flattening rather than new features or behavior. I also made a number of small, unrelated interface-consistency tweaks along the way. Nearly all of these changes are hard to notice unless you compare the old and new versions side by side.

### Checklist


danielelerede-oet:
Wow @eantonini it really looks like you've changed everything. I checked everything carefully but I expect we'll find a few issues here and there while using / modifying the repo, but everything looks fine now

danielelerede-oet:
Hi @eantonini a few comments on the novelties introduced:
- Is it intentional to have the "legend" about the solvers explicit only in the first plot? One can just hover on the bars and see the name of the solver, but at this stage wouldn't it be better to have a common legend on top to have it uniform for all the plots?
image
- I think we can remove this sentence from http://localhost:3000/key-insights?solver=highs#what-benchmark-problems-do-we-have-and-what-are-missing
image
- The new research bar in http://localhost:3000/dashboard/benchmark-problem-set works perfectly
- In the new "Feature Distribution" dashboard, what about having the two fields as "Yes (MILP)" and "No", or "True (MILP)" / "False"? BTW, I find this new section super useful especially for reporting purposes, and I think in the long run we may use it to replace the table at http://localhost:3000/key-insights?solver=highs#what-benchmark-problems-do-we-have-and-what-are-missing
image

eantonini:
Hi @danielelerede-oet, thanks for the detailed check! Addressing these one by one:
- On the landing page, the legend only on the top left panel was intentional. The labels close to the bars help quickly identify the solver they reference to.
- I removed the paragraph at the end of the key insights page.
- In the feature Distribution table, the "Realistic" row now shows "True (MILP)" / "Other".

eantonini (approved):

PR #499, opened 2026-05-04
Website: Panels on benchmark set displaying incorrect number of problems

### LINKED ISSUE #474: Website: Panels on benchmark set displaying incorrect number of problems
The three panels below do not display the correct number of problems, which should be 213.

The total problems in the first panel is 159, 137 in the second, and 274 in the third. Also the height of the bars in the first and second panels should be identical, but they are not.

Image

jacek-oet:
fixes #474 

image

### DIFF
--- a/website/src/components/admin/benchmarks/BenchmarkStatisticsCharts.tsx
+++ b/website/src/components/admin/benchmarks/BenchmarkStatisticsCharts.tsx
@@ -30,7 +30,7 @@ const BenchmarkStatisticsCharts = ({
     );
   }, [metaData]);
 
-  const availabletimeHorizons = ["single", "multi"];
+  const availabletimeHorizons = ["single", "multi", "n/a"];
   const summary = availableModellingFrameworks.map((framework) => {
     const problemClassesMap = new Map<string, number>();
     const applicationsMap = new Map<string, number>();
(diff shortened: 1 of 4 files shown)


eantonini (changes requested):
There are still some issues with the number of problems displayed in these bar charts.

In the first two panels ("By Modeling Framework" and "By Time Horizon"), the sum of all columns is 161, while it is 282 in the third panel.

The labels of the first two panels should also be changed:
- "By Modeling Framework" -> "By Modeling Framework and Problem Class"
- "By Time Horizon" -> "By Modeling Framework and Time Horizon"

In the third panel, the labels on the x-axis should be reversed, i.e., S, M, L (from small to large).

[Edited: I removed the screenshot regarding the missing large problems as it was the intended behavior]

eantonini:
I believe one reason for these issues could be related to the fact that you are displaying what it is called "benchmark problems" in the screenshot below.

I want to see the "benchmark size instances" on all panels.

We are working on a better nomenclature that should fix these ambiguities.

image

jacek-oet:
@eantonini Please review again
image

eantonini:
@jacek-oet Looks great, thanks for fixing this issue.

I just have one last request. Can you sort the modeling frameworks in alphabetical order?

jacek-oet:
@eantonini I updated.
image

eantonini (approved):

PR #532, opened 2026-06-05
Fix categorize_benchmarks.py

danielelerede-oet:
I noticed that `categorize_benchmarks.py` wasn't filtering benchmark metadata files strictly enough. It was matching every YAML file starting with `metadata`, including `metadata_schema.yaml` and `_template_metadata.yaml`, and then trying to process them as benchmark metadata files. This caused the script to fail on `metadata_schema.yaml`, which is a schema file rather than a benchmark metadata file.


siddharth-krishna (approved):
nice catch, thanks!

PR #533, opened 2026-06-05
Campaign creation via CLI

danielelerede-oet:
This PR adds a command-line workflow for creating benchmark campaigns without using the campaign allocation notebooks (based on the notebook proposed in #475).

A new `benchmarks/create_benchmark_campaign.py` CLI tool is added to:

* categorize benchmarks and generate `results/metadata.yaml` automatically,
* select benchmark instances,
* allocate instances across VMs,
* generate OpenTofu campaign files,
* create benchmark campaign directories under `infrastructure/benchmarks/`,
* generate benchmark campaigns from either command-line arguments or configuration files.

### Benchmark selection

* Added support for selecting:

  * all benchmark instances (`--all`),
  * one or more benchmarks (`--benchmark`),
  * specific benchmark instances (`--instance`),
  * metadata size classes (`--size`),
  * metadata names (`--name`).

* Added `--do-not-skip` to include benchmark instances marked with `Skip because` in the metadata.
  By default these instances are excluded.

### Campaign configuration

* Added support for YAML configuration files through:

  ```bash
  --configfile <config.yaml>
  ```

* Added solver selection through `--solver`.

* Default solver set:

  * gurobi
  * highs
  * scip
  * cbc
  * glpk

* Added support for selecting solver versions by year through `--years` (only 2025 is selected by default).

### VM allocation

* Added automatic VM profile selection:

  * S/M instances: `short` profile (`c4-standard-2`, 1 hour timeout)
  * L instances: `long` profile (`c4-highmem-16`, 24 hour timeout)

* Added `--machine-type {short,long}` to override the automatic profile selection.

* Added support for timeout overrides through `--timeout-hours`.

### Documentation

* Added CLI campaign creation documentation to the main README.
* Added configuration file documentation and a complete `config.campaign.default.yaml` example.


danielelerede-oet:
@eantonini following our discussion I have:
- :heavy_check_mark: renamed `include-to-skip` to `do-not-skip`
- :heavy_check_mark: checked the behavior of `years: [2025]` and similar. In the infrastructure, the year identifies the conda environment, not the single solver version, so this would require a larger discussion. I'm not changing anything here at the moment
- :heavy_check_mark: Set up the config file and subsequently updated the README
- :heavy_check_mark: Generate summary csv file in the `infrastructure/` folder (in correspondence to the generated run

benchmarks/config.campaign.default.yaml
```diff
@@ -0,0 +1,52 @@
...
+  all: false # `true` runs all the benchmarks in `results/metadata.yaml`
+
+  benchmark:
+    - benchmark  # example: IESA-Opt-NL / PyPSA-DE
+
+  size: null # Valid values include: S / M / L
+  name: null # Optional filter on benchmark instance names. Example: 10-1h (`name` field in `results/metadata.yaml`)
+
+  # Select specific benchmark instances
+  instance: null # Set to `null` to disable, otherwise example: ["PyPSA-DE:10-1h"], i.e. "benchmark-name:instance-name"
```
eantonini, line 28: Can these options be substituted with just the name or the list of names of the problems (instances) that the user wants to run? So that it becomes consistent with the nomenclature that we wanted to implement.

benchmarks/config.campaign.default.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+  do_not_skip: false # default: `false` automatically excludes benchmark instances marked with 'Skip because:' in the metadata
+
+allocation:
```
eantonini, line 32: For each of the parameters listed under "allocation", can you specify what value the user can use other than the default one? For example, "num_vms" could be an integer? "machine_type" could be a string chosen between "short" and "long"? And so on for the others.

eantonini (commented):
This looks very good, and I like the logic you implemented.

In addition to the individual comments below, I would suggest adding docstrings for each function following NumPy conventions, as well as type hints. These are not required or enforced at the moment, but they would be a useful long-term improvement.

benchmarks/config.campaign.default.yaml
danielelerede-oet: I actually considered this while extending the CLI with config file support. In the end I kept the configuration structure aligned with the CLI as `benchmark` selects one or more benchmark families, `name` filters instance names within those benchmarks, and `instance` allows explicit benchmark instance selections across different benchmarks.  But I agree that once the nomenclature is fully reviewed we can consider these changes

benchmarks/config.campaign.default.yaml
danielelerede-oet: Good idea, initially I didn't want to provide many details, but I think it makes sense. It's now addressed in #545

eantonini (approved):

danielelerede-oet:
@eantonini I've seen you approved this PR, but if #545 is also good I'd rather close this and merge that, so I'd keep this on hold for the moment and not merge it

PR #545, opened 2026-06-19
Running benchmark campaign locally

danielelerede-oet:
This PR builds on #533 and allows benchmark runs on a local machine.

With this setup campaigns are created in:
  * `infrastructure/benchmarks/` (cloud campaigns),
  * `infrastructure/local/benchmarks/` (local campaigns),

### Campaign targets

* Added support for two execution targets:

  * `cloud` (default)
  * `local` (ignores information provided on the machine type)

* Cloud campaigns generate OpenTofu deployment files.

* Local campaigns generate:

  * `local_benchmarks.yaml`
  * `run_local.sh`
  * `campaign_summary.csv`

and can optionally be launched directly from the generated campaign.

Benchmark and solver selection work as in #533.


eantonini (approved):
Looks good to me!

PR #547, opened 2026-06-22
Benchmarks: Update CANOE metadata

idelder:
Updating the CANOE metadata as requested to slightly better reflect the models provided.

If PR into PR is too obnoxious feel free to just copy the changes over. Not worried about contribution or anything

### DIFF
--- a/benchmarks/temoa/metadata.yaml
+++ b/benchmarks/temoa/metadata.yaml
@@ -146,7 +146,7 @@ benchmarks:
       Num. nonzeros: 3225685
 # https://github.com/CANOE
   canoe_national_16d_high:
-    Short description: CANOE national high
+    Short description: A national capacity expansion model for Canada (except territories) with high resolution technology choices for both energy supply and end use demands.
     Modelling framework: TEMOA
     Model name: CANOE
     Version:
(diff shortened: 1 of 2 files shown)


danielelerede-oet:
@idelder thank you for the changes. I'll close the previous PR. Just to confirm: for the benchmarks used for diagnostics does the `Realistic: true` field apply according to you? if they're used for diagnostics only my feeling is that we should classify them as not realistic.

idelder:
Haha I'll hold myself back from a philosophical essay here but I think the model has as much claim to realism as any other in this space. The difference being that while we have real data, etc, we put no stock in being able to predict the future and prioritise dynamic realism instead. How would the energy system realistically respond to a new technology, for example, rather than what would be the realistic end state of that change. And we argue this is the best any model projection of human systems can hope to achieve.

danielelerede-oet:
Thanks @idelder you were pretty convincing ahah I'm flagging all the CANOE instances as realistic then!

danielelerede-oet (approved):

danielelerede-oet:
@siddharth-krishna could you please check what's going on with the Vercel check?

siddharth-krishna:
@danielelerede-oet I think this is a limitation of our Vercel plan, perhaps because the PR is from a fork. Since this PR doesn't touch the website and is not going into the main branch, I think we can ignore the Vercel check and merge it in. Let me know if you don't have permissions to do so.

PR #560, opened 2026-06-30
Add MIPLIB benchmarks

danielelerede-oet:
### Benchmark contributions

This PR adds a selection of MIPLIB 2017 benchmarks from the energy domain. Honestly, I'm against including them because they lack many info required for all other submissions (e.g. the name of the instance, i.e. time and spatial resolution, are just placeholders - indeed, the CI check for name validation is failing because the info is not included in the link). Also, the summary table on the website would be impossible to be updated as for many models we don't even now the framework they belong to.
Anyway, if you deem they are ok to be included, I can infer some more details than those currently included from the descriptions.

### Checklist

- [ ] The benchmark name and size instance name follow the conventions indicated in the template
- [x] I have tested that most of these model instances can be solved to optimality within TO with at least one solver, using the benchmarking infrastructure.

For Benchmark team:
- [ ] Update the summary table in the Key Insights page (if the proposed benchmark(s) is/are adding any new features)
- [x] Test that some solver solves these benchmarks within our timeouts on our infra
- [x] Add CC BY 4.0 license to the metadata file


odow:
> Honestly, I'm against including them because they lack many info required for all other submissions

That's fair. The upcoming MIPLIB 2024 should have more metadata for energy submissions, but it might not have as much as you want. I'm definitely in favor of you keeping this as an energy-modeler-focused set of benchmarks. There is value in being different to MIPLIB. (cc @opt-mucca)

Opt-Mucca:
@odow for MIPLIB 2024 we'd have confirmation that they're "energy instances", but I don't see us ever guaranteeing that time and spatial resolution are available. It's too specific to select instances, e.g., OPF. The information can often be derived from the name or from analysing the instance coefficient matrix too.

danielelerede-oet:
@odow @Opt-Mucca thanks for your observations. We're understanding how to include them in the set, also because those instances fill a gap in MILP benchmarks that we clearly are not able to cover at the moment and would be precious to have.

odow:
It’s also a gap you don’t necessarily have to fill. There is value in being opinionated

eantonini (approved):

Issue #158, opened 2025-04-18
[RFC] What duality gap tolerance should we use while benchmarking MILPs?


siddharth-krishna:
Most solvers allow setting a tolerance for the relative duality gap when solving MILPs to determine when to terminate. (Precisely, if `p` is the primal objective bound (i.e., the incumbent objective value, which is the upper bound for minimization problems), and `d` is the dual objective bound (i.e., the lower bound for minimization problems), then the relative duality gap is defined as `|p - d| / |p|`.)

What tolerance value should we use when benchmarking? Here are the default values used by the solvers:

| Solver | Parameter   | Default |
|--------|-------------|---------|
| HiGHS  | [mip_rel_gap](https://ergo-code.github.io/HiGHS/dev/options/definitions/#mip_rel_gap) | 0.0001  |
| SCIP   | [limits/gap](https://scipopt.org/doc/html/PARAMETERS.php)  | 0       |
| CBC    | [ratioGap](https://github.com/coin-or/Cbc/issues/703#issuecomment-2813395609)    | 0       |
| GLPK   | mipgap      | ??      |
| Gurobi | [MIPGap](https://docs.gurobi.com/projects/optimizer/en/12.0/reference/parameters.html#mipgap)      | 1e-4    |

If no strong opinions, perhaps we go with `1e-4`?

A further complication is that [SCIP](https://scipopt.org/doc/html/PARAMETERS.php) defines `limits/gap` as `|p - d| / min(|d|,|p|)`, and CBC defines `ratioGap` as the "gap between best possible and best known is less than this fraction of larger of two". How can we set a consistent termination condition across all solvers given these slightly different definitions?

(@jajhall, @mattmilten, @pfetsch, @jjhforrest, would love your advice on this matter, thanks in advance!)

mattmilten:
I would recommend using the default values of HiGHS and Gurobi of 1e-4. In practice, it's typically not required to go for a smaller gap.
That the solvers may use slightly different definitions is something we can probably ignore for now.

siddharth-krishna:
Thanks for the comments. None of the smaller benchmarks we've run so far have a small optimal objective values (smallest is `1e4`).

jajhall:
I support @mattmilten's view that a tolerance of 1e-4 should be used.

PR #563, opened 2026-07-03
Review benchmark metrics

danielelerede-oet:
## Summary

This PR improves the collection and reporting of solver quality metrics during benchmark execution by adding support for collecting and reporting the following metrics:

- MIP gap (MILPs)
- LP duality gap (when exposed by the underlying solver API)
- Maximum integrality violation

It also:

- Preserves available solver quality metrics for time-limited runs instead of discarding them.
- Normalizes solver time limit termination to:
  - `Status = TO`
  - `Termination Condition = Timeout`
- Passes benchmark time limits to solver backends while preserving the existing `run_solver.py` interface.
- Improves integer variable detection to avoid unnecessary HiGHS fallback calls for the other solvers.
- Improves timeout handling for LPs where some solver backends cannot return a solution object after reaching the time limit.

## Validation

The implementation was tested locally with both Gurobi and HiGHS.

Verified benchmarks include:

- LP optimal
- LP timeout
- MILP optimal
- MILP timeout

For MILP timeout runs, the benchmark now correctly preserves:

- objective value
- MIP gap
- maximum integrality violation

while reporting the run as a timeout.

## Notes

- LP duality gap is currently only reported when exposed by the underlying solver API.
- GLPK timeout parameter handling has not yet been implemented or verified through the current Linopy interface.


runner/run_solver.py
```diff
@@ -82,7 +82,21 @@ def get_solver(solver_name):
...
 
-    return solver_class(**seed_options.get(solver_name, {}))
+    timeout_options = {
+        "highs": {"time_limit": timeout},
+        "gurobi": {"TimeLimit": timeout},
+        "scip": {"limits/time": timeout},
+        "cbc": {"seconds": timeout},
+        "cplex": {"timelimit": timeout},
+        "xpress": {"timelimit": timeout},
+    }
```
siddharth-krishna, line 92: I'm curious why we want to pass the timeout to the solvers? We already have a mechanism outside the solvers to enforce the timeout using the linux `timeout` command [here](https://github.com/open-energy-transition/solver-benchmark/pull/563/changes#diff-4974f198827ccb6d4ce74824baabe3c3275f6ca743a2b9555af195be73c880eeR254).

siddharth-krishna (commented):
Thanks for this! I have some questions:
- What does the term duality gap mean for LPs? In my head so far, it referred to the MIP gap for MIPs.
- What the current main branch code is calling duality gap is what some people (including [Gurobi](https://docs.gurobi.com/projects/optimizer/en/12.0/reference/parameters.html#mipgap)) call the MIP gap (see also https://github.com/open-energy-transition/solver-benchmark/issues/158). This PR changes the `duality gap` column of the results to be something else, and adds a new MIP gap column. I wonder if this will be confusing to users who are used to the old interpretation? (Unless the old interpretation/nomenclature was wrong, of course.) 
- Also, this change will mean all the results so far are incompatible with any new results since the column `duality gap` has a different meaning. How do we handle this, and will this create any problems?

danielelerede-oet:
> Thanks for this! I have some questions:
> 
> * What does the term duality gap mean for LPs? In my head so far, it referred to the MIP gap for MIPs.
> * What the current main branch code is calling duality gap is what some people (including [Gurobi](https://docs.gurobi.com/projects/optimizer/en/12.0/reference/parameters.html#mipgap)) call the MIP gap (see also #158). This PR changes the `duality gap` column of the results to be something else, and adds a new MIP gap column. I wonder if this will be confusing to users who are used to the old interpretation? (Unless the old interpretation/nomenclature was wrong, of course.)
> * Also, this change will mean all the results so far are incompatible with any new results since the column `duality gap` has a different meaning. How do we handle this, and will this create any problems?

Hey @siddharth-krishna , thanks for the question, I think they're all useful to improve the PR.

The two concepts are related but not identical. For an LP, the duality gap is the difference between the objective values of primal- and dual-feasible solutions. Strong duality implies that this difference is zero at the optimum. However, after looking more carefully at the solver APIs, I agree that `duality_gap` is not an appropriate generic benchmark metric for LPs: solvers do not expose it as a standard output with a uniform definition across algorithms and implementations.

For a MIP, the MIP gap compares the best integer-feasible solution found so far (the incumbent) with the global bound maintained by the branch-and-bound algorithm. Gurobi, for example, defines the relative MIP gap from `ObjVal` and `ObjBound`, and uses the `MIPGap` parameter as a termination tolerance (see
https://docs.gurobi.com/projects/optimizer/en/current/reference/parameters.html#parametermipgap)

Therefore, both quantities describe a remaining optimality gap, but they refer to different bounds and different solution processes.

In this PR, I was using HiGHS's `primal_dual_objective_error` as an LP quality metric. This is one of HiGHS's KKT quality measures, and we cannot currently retrieve a directly equivalent final metric consistently from all solvers. I agree that reporting it as a generic `duality_gap` would therefore be misleading.

I will keep LP primal-dual quality metrics outside the scope of this PR and use `mip_gap` only for MIP problems.

runner/run_solver.py
danielelerede-oet: The solver-level timeout allows the solver to terminate gracefully and return its final state, including the incumbent solution, objective bound, and final MIP gap. With the current implementation we are terminating the entire process externally. If that is the first limit reached, `run_solver.py` cannot retrieve these metrics or write the result. This particularly affects timed-out MIP runs, for which the final MIP gap is one of the main metrics introduced by this PR.

siddharth-krishna: Understood, and that's a nice idea! Thanks

runner/run_solver.py
```diff
@@ -129,55 +143,98 @@ def calculate_integrality_violation(
...
+
+    integrality = getattr(lp, "integrality_", None)
+    col_names = getattr(lp, "col_names_", [])
+
+    if integrality is None:
+        return integer_vars
+
+    for name, var_type in zip(col_names, integrality, strict=False):
+        if var_type == highspy.HighsVarType.kInteger:
+            integer_vars.add(name)
```
siddharth-krishna, line 193: Why are we changing the implementation of how we get the integer variables? Did we test that the new version produces the same results / is correct on some harder benchmarks where we see a nonzero max-integrality-violation? Thanks

runner/run_solver.py
```diff
@@ -206,6 +263,23 @@ def get_reported_runtime(solver_name, solver_model) -> float | None:
...
+        log_text = log_fn.read_text(errors="ignore")
+    except Exception:
+        return False
+
+    timeout_markers = [
+        "Time limit reached",
+        "time limit",
+        "timelimit",
+    ]
+    return any(marker.lower() in log_text.lower() for marker in timeout_markers)
```
siddharth-krishna, line 280: I'm a bit concerned that this might lead to false positives / negatives. Is there some way we can tell this from linopy or the return code? Alternatively, can we grep for these markers in all the log files from the v2 run to test whether this method is robust? Thanks

siddharth-krishna (commented):
Thanks Daniele! I left a few more comments.

PR #567, opened 2026-07-08
State of solvers

eantonini:
This PR adds to the main branch all the files and changes related to the analysis on the state of solvers (Jupyter notebook, figures, blog post).


danielelerede-oet:
@eantonini probably we should also delete the metadata files for the pypsa-de benchmarks. They're not used in the notebooks and have been added to the standard metadata library, so we can get rid of them.

danielelerede-oet (approved):

PR #573, opened 2026-08-03
Runner: update Linopy to 0.9.0

danielelerede-oet:
This PR updates linopy to version 0.9.0 in the test and 2025 benchmark environments, and also restricts Python to 3.12 release, updating the corresponding fixed environment files and adapting solver initialization to the linopy 0.9.0 API by passing solver options through the `options` argument.

I verified locally that:
- no `unexpected keyword argument` errors remain
- the fixed environments use Python 3.12.13 and Linopy 0.9.0
- the original conda channels are preserved

### DIFF
--- a/runner/run_solver.py
+++ b/runner/run_solver.py
@@ -82,7 +82,7 @@ def get_solver(solver_name):
         "xpress": {"miprelgapnotify": mip_gap, "randomseed": 0},
     }
 
-    return solver_class(**seed_options.get(solver_name, {}))
+    return solver_class(options=seed_options.get(solver_name, {}))
 
 
 def is_mip_problem(solver_model, solver_name):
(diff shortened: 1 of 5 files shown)


eantonini (approved):
Looks good, thanks!

runner/run_solver.py
```diff
@@ -82,7 +82,7 @@ def get_solver(solver_name):
         "xpress": {"miprelgapnotify": mip_gap, "randomseed": 0},
     }
 
-    return solver_class(**seed_options.get(solver_name, {}))
+    return solver_class(options=seed_options.get(solver_name, {}))
```
siddharth-krishna, line 85: Does this syntax still work with previous linopy versions that we're using in other envs (< 2025)?

PR #576, opened 2026-08-06
Runner refactoring 1/5: Add pixi as package manager

eantonini:
This PR add `pixi` as a package manager replacing `conda` where applicable.

- Adds a root `pixi.toml` with `runner`, `notebooks`, `benchmarks`, and a shared `dev` environment, replacing `notebooks/pyproject.toml`+`uv.lock` and `benchmarks/environment.yaml`. Per-solver-year conda envs (`runner/envs/*.yaml`) are untouched — that's a later PR in the stack.
- Updates `ci.yaml` to install/run via `pixi` (`setup-pixi` action) instead of bare `pip install`.
- Adds `mypy` to the `dev` feature (available, not yet wired into CI as a gate).
- Updates root/runner/notebooks/benchmarks READMEs for the pixi workflow.

## Notes
- `pandas` for the `benchmarks` env had to go through PyPI rather than conda — conda's `pandas` build pulls a newer `numpy` that conflicts with `highspy==1.9.0`/`linopy` on Python 3.9.

## Test plan
- [x] `pixi install --all` resolves all environments
- [x] `pixi run -e benchmarks python benchmarks/merge_metadata.py` / `validate_urls.py` / `validate_results.py` produce byte-identical output to what's committed
- [x] `pixi run -e runner python -c "import yaml, typer"` succeeds
- [x] `pixi run -e dev` ruff/pytest/mypy/yamale all run


danielelerede-oet:
Thanks Enrico, I have the following comments here:

## `pixi.toml` runner dependencies

Is the `runner` environment expected to run `runner/run_benchmarks.py` directly?

`run_benchmarks.py` imports `HighsVariant` from `run_solver.py`, which imports `linopy` at module load time. However, `linopy` is not included in the `runner` environment or among its resolved packages in `pixi.lock`.

I would therefore expect the following command to fail with `ModuleNotFoundError: linopy`:

```bash
pixi run -e runner python runner/run_benchmarks.py --help
```

I think we should either add `linopy` to the `runner` environment

## `pixi.toml` benchmarks dependencies

Is the `benchmarks` environment intended to support all scripts under `benchmarks/`, including `create_benchmark_campaign.py`?

During campaign creation, that script imports functions from `runner.utils`, which imports `matplotlib`, `humanize`, and `IPython` at module load time. These packages are not included in the `benchmarks` environment so far. If the idea is the one mentioned by me, we should add those dependencies

## `.github/workflows/ci.yaml` pixi setup

The CI currently installs only `dev` and `benchmarks`.

We should probably add some  for the other environments

## `README.md` instructions

`pixi install` only installs the default environment; it does not install all named environments.

Should this be:

```bash
pixi install --all
```

This would also be consistent with the command reported in the PR test plan.

## `.github/workflows/ci.yaml` — `validate_urls.py` error message

Since this script is now executed through the `benchmarks` pixi environment, should the suggested command be updated as well?

```bash
pixi run -e benchmarks python tests/validate_urls.py
```

The currently displayed standalone `python` command may fail in a clean checkout where the dependencies are only provided by pixi.

## `.github/workflows/ci.yaml` and `merge_metadata.py` error message

Same issue here: should the command shown to the user match the new pixi-based workflow?

```bash
pixi run -e benchmarks python benchmarks/merge_metadata.py
```

Otherwise, the CI executes one command but recommends a different one when the generated metadata is outdated.

eantonini:
Thanks for the review, @danielelerede-oet. Below are the fixes following your comments:

- `linopy` added to `runner`'s pypi-deps, and `runner` added to CI's pixi `environments:`. Both will collide (trivially) with PR 2/5, which adds these too for its test suite. Expect a one-line conflict on the next rebase.
- `matplotlib`/`humanize`/`ipython` added to `benchmarks` (matplotlib pinned to `3.7.*`), plus `from __future__ import annotations` in `runner/utils.py`. This becomes dead weight once PR 3/5's module split happens.
- README (`pixi install --all`) and CI's suggested-fix messages (`pixi run -e benchmarks python ...`).

Re-verified `pixi install --all`, pre-commit, and all CI steps this PR touches.

danielelerede-oet:
Thanks @eantonini . I rechecked the latest changes and they address my comments: the missing dependencies have been added, the `runner` environment is now installed in CI, and the README and suggested CI commands have been corrected. The CI has passed too, so it's good to go from my side.

danielelerede-oet (approved):

PR #578, opened 2026-08-06
Runner refactoring 2/5: Characterization tests for runner/utils.py, run_solver.py, run_benchmarks.py

eantonini:
This PR locks in current behavior of the runner scripts before the actual refactor splits them into a module package.

- Adds `runner/tests/` (conftest.py + test_utils.py + test_run_solver.py + test_run_benchmarks.py), 106 tests covering: `calculate_sgm`, `is_solved`, `compute_summary_results`, `allocate_vms_greedy`, `allocate_benchmarks`, `load_benchmark_metadata`; `HighsVariant`, `is_mip_problem`, `get_duality_gap`, `get_reported_runtime`, `calculate_integrality_violation`, `get_solver`'s seed-options wiring; `csv_record`, `write_csv_headers`/`write_csv_row` round-trip, `parse_memory`, `get_conda_package_versions`, `download_benchmark_file`, `benchmark_solver`'s timeout/OOM/error branches.
- Note: several functions the original issue attributed to `utils.py` (`csv_record`, `parse_memory`, `get_conda_package_versions`, `download_benchmark_file`, `benchmark_solver`) actually live in `run_benchmarks.py` today — tested there instead, matching the real current layout.
- Wires `pixi run -e runner pytest runner/tests -v` into CI, right after metadata-schema validation.
- Extends the `runner` pixi feature with `matplotlib`, `numpy`, `ipython`, `linopy` — needed to import `utils.py`/`run_solver.py`/`run_benchmarks.py` at all (they pull these in at module level even for non-plotting/non-solving functions). None of this is needed by the CLI in production; it's exactly the coupling Refactor PR 3's module split removes, so expect this to shrink back out then.

## Notable finding
Characterization testing surfaced a real pre-existing bug: `benchmark_solver` crashes with an uncaught `IndexError` (not `ValueError`) if the wrapped subprocess produces completely empty stderr, since `parse_memory` does `output.splitlines()[-1]` with no guard for empty input. Documented as a test locking in current behavior rather than fixed here — out of scope for a characterization-only PR.

## Test plan
- [x] `pixi run -e runner pytest runner/tests -v` — 106 passed
- [x] No production-code diff introduced (only pre-existing bot-formatting noise already on the branch)
- [x] `ruff check`/`ruff format --check` clean on new test files

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,7 +26,7 @@ jobs:
 
       - uses: prefix-dev/setup-pixi@v0.10.1
         with:
-          environments: dev benchmarks
+          environments: dev benchmarks runner
 
       - name: Check code formatting
         run: |
@@ -69,6 +69,9 @@ jobs:
 
       - name: Validate metadata schema
         run: pixi run -e dev yamale -s benchmarks/metadata_schema.yaml results/metadata.yaml
+
+      - name: Run runner unit tests
+        run: pixi run -e runner pytest runner/tests -v
 
       - name: Set FORK_PR environment variable
         if: ${{ (github.event.pull_request.head.repo.fork == true || github.actor == 'dependabot[bot]') && github.event_name == 'pull_request' }}
--- a/pixi.lock
+++ b/pixi.lock
@@ -86,6 +86,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/noarch/pygments-2.19.2-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/pysocks-1.7.1-pyha55dd90_7.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/pytest-8.4.1-pyhd8ed1ab_0.conda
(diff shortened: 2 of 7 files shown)


danielelerede-oet:
My only concern here is about `test_empty_stderr_raises_indexerror`. Since the uncaught `IndexError` is a known bug rather than intended behavior, would it be better to test the desired result (`memory=None`) and mark it as `xfail(strict=True)`? I would review the actual module separation and removal of the temporary `matplotlib`/`IPython`/`linopy` coupling in the subsequent PR.

eantonini:
Thanks, @danielelerede-oet. I renamed to `test_empty_stderr_leaves_memory_none`, asserts `memory is None` (the desired behavior) wrapped in `@pytest.mark.xfail(strict=True)`. Passes as an expected failure today; `strict=True` means it'll hard-fail if the bug is ever fixed without updating this test.

danielelerede-oet:
Thanks @eantonini , the updated test now expresses the desired behavior. Approved from my side.

danielelerede-oet (approved):

PR #579, opened 2026-08-07
Runner refactoring 3/5: Split runner/utils.py into a modular, config-driven package

eantonini:
This PR splits runner/utils.py into a modular, config-driven package

- Replaced the monolithic `runner/utils.py` (and `run_solver.py`) with a `runner/utils/` package split by concern: `config`, `env`, `execution`, `metadata`, `campaign`, `results`, `analyze`, `plot`, `cloud`, `monitor`, and solver dispatch (`solver.py` + a `solvers/` package with one auto-discovered adapter module per solver).
- Moved the solver registry, tuning options, and version/eligibility guards out of hardcoded Python into `runner/config/*.yaml` (`solvers.yaml`, `solver_configurations.yaml`, `eligibility_rules.yaml`), driven by a small generic rule engine (`in`/`not_in`/`eq`/`gte`/`lte`) instead of ad hoc conditionals — adding a solver, a tuning configuration, or a new runtime guard is now a config edit, not new code.
- Renamed "Benchmark"/"Instance"/"Size"-instance nomenclature to "Problem" throughout code, CSVs, docs, and notebooks, reflecting that metadata.yaml has been flat (one problem per model+size) since #481. Historical CSVs still load correctly via explicit backward-compat handling in `analyze.load_results`.
- `run_solver.py`'s and `run_benchmarks.py`'s logic now live in `runner.utils.solver` and `runner.utils.orchestrator` respectively, invoked via `python -m` for reliable package-relative imports regardless of subprocess wrapping (`systemd-run`, `/usr/bin/time`, `conda run`).
- Redesigned `benchmarks/create_benchmark_campaign.py`'s selection CLI (`--problem`/`--size`) and updated it, six notebooks' imports, and the READMEs for the new schema and nomenclature.
- Migrated the test suite to one file per module (124 tests), including new coverage for the config registry, eligibility rules, and orchestrator that didn't exist before; fixed a latent `StatisticsError` crash in the run loop (mean/stddev computed against the requested iteration count instead of runs actually completed) found while writing tests.


runner/utils/campaign.py
```diff
@@ -0,0 +1,173 @@
...
+    return allocation, weights
+
+
+def allocate_problems(
+    problems_df: pd.DataFrame,
+    weight_col: str,
+    num_vms: int,
+    machine_type: str = "c4-standard-2",
+    zone: str = "us-central1-a",
+    solvers: str | None = None,
```
danielelerede-oet: The `benchmarks` Pixi environment is pinned to Python 3.9, but this module uses PEP 604 annotations without `from __future__ import annotations`.

This currently prevents the campaign generator from running. I reproduced it with:

```bash
pixi run -e benchmarks python \
  benchmarks/create_benchmark_campaign.py \
  --configfile benchmarks/config.campaign.default.yaml \
  --skip-prepare
```

which fails while importing `runner.utils.campaign` with:

```text
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

The same pattern appears in several new `runner.utils` modules, including `metadata.py`, which is also imported by the campaign generator. 

Could we add `from __future__ import annotations` to the modules that must support Python 3.9, or otherwise avoid these annotations? An actual campaign-generation test in the `benchmarks` environment would catch this.

runner/config/solvers.yaml
```diff
@@ -0,0 +1,115 @@
+# Solver registry consumed by runner/utils/config.py.
```
danielelerede-oet: This file replaces `runner/solvers.yaml`, but `runner/benchmark_all.sh` still opens the deleted path:

```python
config = yaml.safe_load(open('./runner/solvers.yaml'))
```

Could we update `benchmark_all.sh` to use `runner/config/solvers.yaml` and update the remaining references in `runner/README.md` and the root README as well?

runner/utils/results.py
```diff
@@ -0,0 +1,235 @@
...
+    Has no "Size" column: that held the specific instance name within a
+    problem (e.g. a resolution/cluster-count like "100-12h" for a
+    multi-instance problem family), which is always "default" today since
+    metadata.yaml is now one row per model+size combination already (see
+    `metadata.load_problems`) -- nothing left to disambiguate. Historical
+    CSVs still have it; `analyze.load_results` knows how to read those.
+    """
+    record = OrderedDict(
+        [
+            ("Problem", kwargs.get("problem_id")),
```
danielelerede-oet: Changing the generated csv schema from `Benchmark,Size,Solver,...` to `Problem,Solver,...` is not currently propagated to all consumers.

In particular:

- `tests/validate_results.py` still reads and groups by `Benchmark` and `Size`;
- `website/src/utils/results.ts` still reads `data["Benchmark"]` and `data["Size"]`;
- the website solver types, labels, and colours do not recognize names such as `highs-default` and `gurobi-default`;
- the CI `awk` check still treats column 3 as the solver and column 6 as the status, whereas in the new schema they are `Solver Version` and `Termination Condition`.

The backward-compatibility handling in `analyze.load_results()` only covers that Python analysis path. Could we either update all CSV consumers in this PR or preserve the existing output schema until they are migrated?

benchmarks/create_benchmark_campaign.py
```diff
@@ -455,50 +389,29 @@ def build_parser(defaults: dict | None = None) -> argparse.ArgumentParser:
         "--all",
         action="store_true",
         default=argparse.SUPPRESS,
-        help="Select all benchmark instances.",
+        help="Select all problems.",
     )
     selection.add_argument(
-        "--benchmark",
+        "--problem",
```
danielelerede-oet: The new selection interface accepts only `all`, `problem`, `size`, and `do_not_skip`, but `benchmarks/config.campaign.default.yaml` still defines the removed `benchmark`, `name`, and `instance` selectors.

Those keys are silently ignored by `flatten_config()`, so the supplied configuration resolves to `all=False` and `problem=None` and then fails with:

```text
Select exactly one mode: --all or --problem.
```

Could we migrate the selection section of `config.campaign.default.yaml` to the new schema and add a test that loads the shipped configuration file?

danielelerede-oet:
Hi @eantonini this was a tough one. The modular split itself is clear, and I ran the new runner test suite successfully.
Nonetheless, I have identified a few points worth to be checked: these should be integration blockers rather than issues with the module split itself. The current checks remain green because the runner tests use the newer `runner` environment, the campaign generator is not executed in CI, the sample benchmark uses the special `tests` branch, and the website build consumes the already committed historical CSV rather than a newly generated one.

PR #580, opened 2026-08-07
Runner refactoring 4/5: Unify run_benchmarks.py + benchmark_all.sh into a single runner.benchmark CLI

eantonini:
This PR unifies `run_benchmarks.py` + `benchmark_all.sh` into a single `runner.benchmark` CLI

- Replaced `runner/run_benchmarks.py` (single-year argparse CLI) and `runner/benchmark_all.sh` (bash multi-year loop + per-solver conda env setup) with one Typer CLI, `runner/benchmark.py` (`python -m runner.benchmark`). It loops over years, creates any missing per-solver-year conda envs (`env.ensure_solver_envs_installed`, new), and delegates to `orchestrator.run_benchmark` — no more borrowing a solver env's own Python for the orchestrator process itself, since pixi now provides the CLI's dependencies directly.
- Both `--years` and `--solver-configurations` (renamed from `--solvers` for consistency with the solver vs. solver-configuration distinction used elsewhere) default from config rather than a hardcoded list: `config.get_all_registered_years()` reads the year set straight from `solvers.yaml`, so a new solver-version year needs no CLI code change.
- **Found and fixed a real, pre-existing bug**: `solvers.yaml` never had a `year: "tests"` entry, so CI's `-y tests` smoke-test step silently ran zero solvers and "passed" vacuously (an empty results CSV trivially satisfies the awk check that follows it) — on `main` too, not just this branch. Added a proper `tests:` block to `solvers.yaml` mapping to `runner/envs/benchmark-tests.yaml`'s actual pinned versions, taught `env.get_registered_solver_versions` to read it, and hardened `config`'s `gte`/`lte` eligibility operators against the non-numeric `"tests"` pseudo-year (they previously raised `ValueError`).
- Updated every caller of the old scripts to the new CLI: the Dockerfile (now installs pixi; entrypoint runs `runner.benchmark`), `infrastructure/startup-script.sh` (GCE VM provisioning), `create_benchmark_campaign.py`'s generated `run_local.sh`, the CI workflow, and all docs (root `README.md`, `runner/README.md`, `infrastructure/README.md`, `docs/Metrics_and_methodology.md`).
- Trimmed `pixi.toml`'s `runner` feature: `matplotlib`/`humanize` turned out to be `plot.py`-only (notebook-side, never imported from `runner/`) — removed `matplotlib`; `numpy`/`ipython`/`linopy` stay since `analyze.py`/`solver.py` are imported directly by the test suite.
- Added CLI test coverage (`test_benchmark.py`, `typer.testing.CliRunner`) plus new tests for `env.ensure_solver_envs_installed` and `config.get_all_registered_years`, including three regression guards for the "tests" bug specifically: an end-to-end CLI run against the real (unmocked) solver registry, a drift check between `solvers.yaml`'s `tests:` block and `benchmark-tests.yaml`'s actual pinned versions, and a subprocess-level smoke test of the real `python -m runner.benchmark` invocation contract.
- Fixed two additional CI regressions surfaced by the "tests" pseudo-year fix (which, for the first time, made the sample-benchmarks step actually write real rows instead of silently writing zero): the `awk` success check used column indices (`$3`/`$6`) from the old CSV schema instead of the new one (`$2`/`$5`), and `solver.py`'s MILP integrality-violation calculation still assumed linopy's `Solution.primal` was a name-indexed `pandas.Series`, when linopy 0.9 changed it to a label-indexed `numpy.ndarray` (`AttributeError: 'numpy.ndarray' object has no attribute 'loc'`).
- Restricted CI's sample-benchmark run to `cbc`, `highs`, `scip` (dropping `glpk`), sidestepping an unrelated, non-blocking upstream linopy/GLPK dual-value-parsing bug rather than continuing to tolerate its noisy traceback.


runner/benchmark.py
```diff
@@ -0,0 +1,105 @@
...
+        try:
+            run_benchmark(
+                problems_yaml_path,
+                resolved_solver_configurations,
+                year=year,
+                reference_interval=ref_bench_interval,
+                append=append or index > 0,
+                run_id=resolved_run_id,
+            )
+        except Exception as e:
```
danielelerede-oet: The CLI currently catches each year-level exception, continues, and eventually exits successfully even if one or all requested years failed. This conflicts with `infrastructure/startup-script.sh`, which relies on `BENCHMARK_EXIT_CODE` to determine whether the benchmark succeeded and whether results should be uploaded.

We  should preserve the continue-on-error behavior, but collect the failed years and exit with a non-zero status after all years have been processed maybe? It would also be useful to fail explicitly when `registered_versions` is empty, since an unsupported year or configuration currently produces a header-only CSV and still reports success.

.github/workflows/ci.yaml
```diff
@@ -111,11 +111,15 @@ jobs:
...
+          pixi run -e runner python -m runner.benchmark --years tests \
+            --solver-configurations cbc \
+            --solver-configurations highs \
+            --solver-configurations scip \
+            results/metadata_subset.yaml
           echo ""
           cat results/benchmark_results.csv
-          if awk -F',' 'NR>1 && $3 != "glpk-default" {if ($6 != "ok") exit 1}' results/benchmark_results.csv; then
-            echo "All non-GLPK benchmarks ran successfully."
+          if awk -F',' 'NR>1 {if ($5 != "ok") exit 1}' results/benchmark_results.csv; then
```
danielelerede-oet: The corrected status column is right, but this check still succeeds when the CSV contains only its header: no `NR > 1` action is executed, so `awk` exits with status 0.
Maybe we should also require at least one result row? For example:

`awk -F',' 'NR > 1 { rows++; if ($5 != "ok") failed=1 } END { exit(rows == 0 || failed) }' results/benchmark_results.csv`

Ideally, the check could also verify that all expected solver configurations produced a row.

danielelerede-oet:
@eantonini I've added a few comments in the files. In addition, `infrastructure/README.md` still mentions  `benchmark_all.sh` though it's removed by this PR

PR #581, opened 2026-08-07
Runner refactoring 5/5: Migrate per-solver-year conda environments to pixi

eantonini:
This PR replaces every per-solver-year conda environment (`runner/envs/*.yaml`) with a self-contained pixi manifest (`runner/envs/<solver>-<year>/pixi.toml` + `pixi.lock`), and removes conda from the project entirely.

- Add 21 `runner/envs/<name>/pixi.toml` manifests (one per solver-year, isolated from the root `pixi.toml` so each can resolve/fail independently), replacing the old loose + `-fixed` conda YAMLs and `generate_fixed_envs.sh`.
- Rewrite `runner/utils/env.py` to drive `pixi list`/`pixi install` instead of `conda list`/`conda env create`; `ensure_solver_envs_installed` now relies on `pixi install`'s own idempotency rather than a hand-rolled "does this env already exist" check.
- Switch `runner/utils/execution.py`'s solver subprocess from `conda run -n <env>` to `pixi run --manifest-path runner/envs/<env>`.
- Rename `config.get_conda_package_name` → `get_package_name` now that the lookup isn't conda-specific.
- `runner/Dockerfile`: base image `continuumio/miniconda3` → `debian:bookworm-slim`; `infrastructure/startup-script.sh`: drop the Miniconda install/ToS-accept block, since pixi's own installer already covers everything.
- Fix two real linopy 0.9 compatibility bugs surfaced while verifying real end-to-end solves against the new envs: `Solution.primal`/`.dual` are label-indexed `np.ndarray`s (not name-indexed `pd.Series`), and linopy's deprecated file-based solve path never sets its internal variable count, so `Solution.primal` is always empty for that path — `calculate_integrality_violation` now handles both.
- One env needed a genuine packaging workaround: `xpress`'s conda channel has never published a working Linux build (missing `intel-openmp` for linux-64, an upstream FICO packaging bug), so `benchmark-xpress-2025` installs `xpress` from PyPI instead.
- Update `README.md`, `runner/README.md`, `runner/SOLVERS.md`, `docs/Criteria_and_instructions.md`, `infrastructure/README.md`, and CI to reflect pixi throughout, and drop the now-redundant `--solver-configurations` exclusions in CI.

## Test plan
- [x] Each of the 21 new envs verified via `pixi install --manifest-path runner/envs/<name>` and a real `pixi run ... python3 -c "import <package>"` smoke import.
- [x] Full unit test suite updated and passing (`test_env.py`, `test_execution.py`, `test_config.py`, `test_solver.py`) for the pixi-based env/execution paths and the two linopy fixes, including regression tests for the empty-primal case.
- [x] Ran a real end-to-end MILP solve through the CLI against the new envs to confirm the linopy bugs above and validate the fix.
- [x] CI's sample-benchmark job passes with `cbc`/`highs`/`scip` against the `benchmark-tests` pixi env.


runner/Dockerfile
```diff
@@ -6,26 +6,30 @@
...
-# Per-solver-year conda environments are created at runtime by
-# runner/benchmark.py. To cache them across runs, use a named volume:
+# Per-solver-year pixi environments (runner/envs/<env>/) are installed at
+# runtime by runner/benchmark.py. To cache them across runs, use a named
+# volume:
 #   docker run --rm \
 #     -v $(pwd)/results:/solver-benchmark/results \
-#     -v solver-conda-envs:/opt/conda/envs \
-#     solver-benchmark-runner --solver-configurations highs --years 2025 results/metadata.yaml
+#     -v solver-pixi-envs:/solver-benchmark/runner/envs \
```
danielelerede-oet: Mounting the volume over the whole `runner/envs` directory also hides the `pixi.toml` and `pixi.lock` files copied into the image.

We could probablu create a separate cache directory or detached environments? The corresponding examples in `runner/README.md` should then use that cache location as well.

runner/utils/solver.py
```diff
@@ -122,16 +122,28 @@ def calculate_integrality_violation(
...
+    `primal_values` can legitimately be shorter than `integer_var_labels`
+    needs: linopy's deprecated file-based solve path (`Solver.solve_problem`,
+    see this module's docstring) never initializes its internal variable
+    count, so it always builds `Solution.primal` as a zero-length array
+    regardless of the problem's real size. Labels beyond its length are
+    dropped rather than raising, since there's no primal value to check
+    them against either way.
     """
-    p = primal_values[integer_var_labels]
+    in_bounds = integer_var_labels[integer_var_labels < len(primal_values)]
```
danielelerede-oet: This avoids the exception, but it does not restore the maximum integrality violation metric.

This avoids the exception, but it does not restore the maximum integrality violation metric. 

Could we recover the primal values from the generated solution file or the native solver model, or switch away from the deprecated linopy path that returns the empty array? If not all requested integer labels are available, I think the whole metric should remain unavailable rather than being calculated from a partial subset.

runner/utils/env.py
```diff
@@ -170,29 +150,19 @@ def ensure_solver_envs_installed(
...
+        env_dir = _ENVS_DIR / env_name
+        if not (env_dir / "pixi.toml").exists():
+            print(f"WARNING: No pixi manifest found for env {env_name}, skipping")
             continue
 
-        print(f"Creating conda env {env_name} from {env_yaml.name}...")
+        print(f"Ensuring env {env_name} is installed...")
         result = subprocess.run(
-            ["conda", "env", "create", "-q", "-f", str(env_yaml), "-y"],
+            ["pixi", "install", "--manifest-path", str(env_dir)],
```
danielelerede-oet: The committed lockfiles are intended to make benchmark environments reproducible, but `pixi install` updates the lockfile automatically when it considers it out of date. `pixi run` in `runner/utils/execution.py` can do the same.

Could we add `--locked` to both commands, or set `PIXI_LOCKED`, so a manifest/lock mismatch fails explicitly instead of silently re-resolving dependencies during a benchmark run?

runner/README.md
```diff
@@ -141,25 +136,13 @@ python -m runner.utils.solver <solver_configuration> <input_file> <solver_versio
 
 ```bash
 # Test HiGHS (from the repo root)
-conda activate benchmark-highs-2024
-python -m runner.utils.solver highs-default runner/benchmarks/pypsa-eur-elec-op-2-1h.lp 1.10.0
+pixi run --manifest-path runner/envs/benchmark-highs-2024 python -m runner.utils.solver highs-default runner/benchmarks/pypsa-eur-elec-op-2-1h.lp 1.10.0
```
danielelerede-oet: These example version arguments do not match the referenced environments. `benchmark-highs-2024` contains HiGHS 1.9.0, not 1.10.0, and `benchmark-scip-2024` contains SCIP 9.2.0, not 9.2.2.

danielelerede-oet (changes requested):

